PR #34, opened 2017-05-20
Added "...ToString" functions for all SPIRV/1.2 enums.

LukasBanana:
I created three new header files with `...ToString` functions for all SPIRV/1.2 enums:
`spirv_strings.h`, `spirv_strings.hpp`, and `spirv_strings.hpp11`.

They can be used like this in C++11:
```cpp
// Defines the function implementations
// (similar to https://github.com/nothings/stb/blob/master/stb_image.h)
#define SPIRV_STRINGS_IMPLEMENT
#include <spirv/1.2/spirv_strings.hpp11>

/* ... */

const char* s1 = spv::OpToString(spv::Op::OpName); // returns "OpName"
const char* s2 =  spv::SourceLanguageToString(spv::SourceLanguage::GLSL); // returns "GLSL"
```

Or like this in C:
```c
#define SPIRV_STRINGS_IMPLEMENT
#include <spirv/1.2/spirv_strings.h>

/* ... */

const char* s1 = SpvOpToString(SpvOpName); // returns "OpName"
const char* s2 =  SpvSourceLanguageToString(SpvSourceLanguageGLSL); // returns "GLSL"
```

I needed this for a custom disassembler and SPIRV debugging functionality.

### DIFF
--- a/include/spirv/1.2/spirv_strings.h
+++ b/include/spirv/1.2/spirv_strings.h
@@ -1,0 +1,1284 @@
+/*
+** Copyright (c) 2014-2017 The Khronos Group Inc.
+** 
+** Permission is hereby granted, free of charge, to any person obtaining a copy
+** of this software and/or associated documentation files (the "Materials"),
+** to deal in the Materials without restriction, including without limitation
+** the rights to use, copy, modify, merge, publish, distribute, sublicense,
+** and/or sell copies of the Materials, and to permit persons to whom the
+** Materials are furnished to do so, subject to the following conditions:
+** 
+** The above copyright notice and this permission notice shall be included in
+** all copies or substantial portions of the Materials.
+** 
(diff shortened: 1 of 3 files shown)


dneto0:
The preferred way to do this kind of thing is to generate your own private headers from the JSON grammar files.
That minimizes the maintenance burden as extensions are added.

For example, SPIRV-Tools generates various tables from the grammar files for the core instruction set and extended instruction sets.
The processing is by in https://github.com/KhronosGroup/SPIRV-Tools/blob/master/utils/generate_grammar_tables.py 
It's invoked from https://github.com/KhronosGroup/SPIRV-Tools/blob/master/source/CMakeLists.txt#L23

A simpler example is how SPIRV-Tools can be used to generate syntax tables from Vim syntax colouring.
See https://github.com/KhronosGroup/SPIRV-Tools/blob/master/source/CMakeLists.txt#L59

LukasBanana:
Thanks for the reply. I'll try using the SPIRV-Tools.

PR #37, opened 2017-06-19
Add SPIR-V tokens that related with AMD extensions

amdrexu:


amdrexu:
As we discussed, please help me to add those tokens to SPIR-V headers and propagate them to all SPIR-V versions. Thank you very much.

johnkslang:
This is now applied internally in the Khronos source. Changes to this file and headers, etc., will show up next time headers are updated.

amdrexu:
Thanks for your kind help.

johnkslang:
Note that many of the final entries in lists added ended with `,` and that is not legal according to the JSON processor SPIR-V is using.

PR #42, opened 2017-07-26
Add new tokens defined by SPV_EXT_shader_stencil_export

amdrexu:


amdrexu:
@johnkslang Could you help to update the header file based on the spec: https://www.khronos.org/registry/spir-v/extensions/EXT/SPV_EXT_shader_stencil_export.html. Thanks very much.

johnkslang:
This has been pulled into the Khronos-private source. Will be visible here on next header update.

amdrexu:
Good news. Thank you.

PR #57, opened 2018-02-13
Allow SubgroupSize and SubgroupLocalInvocationId with SubgroupBallotKHR

antiagainst:
As specified in https://www.khronos.org/registry/spir-v/extensions/KHR/SPV_KHR_shader_ballot.html:

```
 (Add the SubgroupBallotKHR capability to SubgroupSize.)
 (Add the SubgroupBallotKHR capability to SubgroupLocalInvocationId.)
```

### DIFF
--- a/include/spirv/1.0/spirv.core.grammar.json
+++ b/include/spirv/1.0/spirv.core.grammar.json
@@ -5005,7 +5005,7 @@
         {
           "enumerant" : "SubgroupSize",
           "value" : 36,
-          "capabilities" : [ "Kernel" ]
+          "capabilities" : [ "Kernel", "SubgroupBallotKHR" ]
         },
         {
           "enumerant" : "SubgroupMaxSize",
@@ -5030,7 +5030,7 @@
         {
           "enumerant" : "SubgroupLocalInvocationId",
           "value" : 41,
-          "capabilities" : [ "Kernel" ]
+          "capabilities" : [ "Kernel", "SubgroupBallotKHR" ]
         },
         {
           "enumerant" : "VertexIndex",
--- a/include/spirv/1.1/spirv.core.grammar.json
+++ b/include/spirv/1.1/spirv.core.grammar.json
@@ -5153,7 +5153,7 @@
         {
           "enumerant" : "SubgroupSize",
           "value" : 36,
-          "capabilities" : [ "Kernel" ]
+          "capabilities" : [ "Kernel", "SubgroupBallotKHR" ]
         },
         {
           "enumerant" : "SubgroupMaxSize",
@@ -5178,7 +5178,7 @@
         {
           "enumerant" : "SubgroupLocalInvocationId",
           "value" : 41,
-          "capabilities" : [ "Kernel" ]
+          "capabilities" : [ "Kernel", "SubgroupBallotKHR" ]
         },
         {
           "enumerant" : "VertexIndex",
--- a/include/spirv/1.2/spirv.core.grammar.json
+++ b/include/spirv/1.2/spirv.core.grammar.json
@@ -5210,7 +5210,7 @@
         {
           "enumerant" : "SubgroupSize",
           "value" : 36,
-          "capabilities" : [ "Kernel" ]
+          "capabilities" : [ "Kernel", "SubgroupBallotKHR" ]
         },
         {
           "enumerant" : "SubgroupMaxSize",
@@ -5235,7 +5235,7 @@
         {
           "enumerant" : "SubgroupLocalInvocationId",
           "value" : 41,
-          "capabilities" : [ "Kernel" ]
+          "capabilities" : [ "Kernel", "SubgroupBallotKHR" ]
         },
         {
           "enumerant" : "VertexIndex",
--- a/include/spirv/unified1/spirv.core.grammar.json
+++ b/include/spirv/unified1/spirv.core.grammar.json
@@ -5229,7 +5229,7 @@
         {
           "enumerant" : "SubgroupSize",
           "value" : 36,
-          "capabilities" : [ "Kernel" ]
+          "capabilities" : [ "Kernel", "SubgroupBallotKHR" ]
         },
         {
           "enumerant" : "SubgroupMaxSize",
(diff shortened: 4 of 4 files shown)


johnkslang:
Right, the correct details are in the extension. The extension is not incorporated into the spec or headers yet. You have to look at the extension. Only the tokens should get reserved, and even doing that has turned out to be questionable, because it keeps fooling readers into thinking the functionality is present, when it is not.

A big part of the unified headers/specification was to address this "tricking" of readers in thinking stuff is in the spec. when it really is not.  In the non-unified world, what always happened when I incorporate an extension was that the list of extensions went away, while the list of capabilities showed up.

Perhaps you want to list the extension instead of listing the capabilities?  

We need to go more in the direction of being clear something is an extension and not in the spec, instead of going the direction of putting even more of an unincorporated extension into the spec.

antiagainst:
> Right, the correct details are in the extension. The extension is not incorporated into the spec or headers yet.

But we have already put the `Subgroup{Eq|Ge|Gt|Le|Lt}MaskKHR` builtins and other instructions (like `OpSubgroupBallotKHR`) from the `SPV_KHR_shader_ballot` extension into the spec and header?

> You have to look at the extension. Only the tokens should get reserved, and even doing that has turned out to be questionable, because it keeps fooling readers into thinking the functionality is present, when it is not.

Sorry, I was not aware of this side effect; I was purely thinking from tooling's perspective. Agreed. You are right. Although from tooling's perspective, having a single source of truth for (ideally) everything is quite preferable, it blurs the boundary between core spec and extensions.

I think this also kind of solves my previous question. Those symbols are placed into the spec to get reserved.

> A big part of the unified headers/specification was to address this "tricking" of readers in thinking stuff is in the spec. when it really is not.

+1

> In the non-unified world, what always happened when I incorporate an extension was that the list of extensions went away, while the list of capabilities showed up.

> Perhaps you want to list the extension instead of listing the capabilities?

I'd think I need to provide a little bit explanation of why I'm doing like this way. It's because we have `Subgroup{Eq|Ge|Gt|Le|Lt}MaskKHR` in the grammar/spec, with their required capabilities directly specified as `SubgroupBallotKHR`. I think `SubgroupSize` and `SubgroupLocalInvocationId`, although not introduced by `SPV_KHR_shader_ballot`, is parallel to `Subgroup{Eq|Ge|Gt|Le|Lt}MaskKHR`; so it should also be fine to have `SubgroupBallotKHR` as their enabling capability. But it's just kind of my reasoning plus guessing, please definitely correct me if I get something wrong.

I'm not sure how to specify the `SPV_KHR_shader_ballot` extension here. Attaching an `extension` field to  `SubgroupSize`/`SubgroupLocalInvocationId` will make it look like `SubgroupSize`/`SubgroupLocalInvocationId` is available only in `SPV_KHR_shader_ballot`, which is of course incorrect.

I actually does not have a strong stand to have this pull request merged. Closing this and fixing the tooling (SPIRV-Tools) is also an acceptable solution.

> We need to go more in the direction of being clear something is an extension and not in the spec, instead of going the direction of putting even more of an unincorporated extension into the spec.

Agreed. The spec should be clear about the boundaries: what functionalities are in the core spec and what in extensions.

IMHO, there is also a difference between the spec and the grammar: I think the grammar is more of a source of truth that tools can rely on to generate various parts. The spec itself is a consumer of the grammar; it only takes what are suitable to be in the spec and discards the rest. SPIRV-Tools is another consumer of the grammar; it automatically gets lots of facts from the grammar.

Generally it's fine (and preferable probably) to have dedicated grammar files for extensions. The problem rises when the extension want to **modify** what the core spec says, like what we see here: `SPV_KHR_shader_ballot` is trying to modify `SubgroupSize` and `SubgroupLocalInvocationId`. There is just no way to represent such facts right now. Adding it into the core grammar is the easy approach, but it does bring the question of where is the cut. Probably KHR extensions have some privilege here? I'm not sure though.

johnkslang:
I think this is a good discussion.

If the capability is currently in the spec inconsistently, I agree that is wrong, and should be fixed.

It could be that the new unified way helps you: Because now we explicitly list something is "Reserved", we don't have to rely on the list of extensions being the flag.  Still, I'm unclear that the core spec. should be including details of an extension before it has been agreed they belong in core.

antiagainst:
> If the capability is currently in the spec inconsistently, I agree that is wrong, and should be fixed.

Yep. In the spec and grammar, `SubgroupBallotKHR` is already listed as the enabling capability for all other builtins/instructions in `SPV_KHR_shader_ballot`. This PR fixes `SubgroupSize` and `SubgroupLocalInvocationId`.

> It could be that the new unified way helps you: Because now we explicitly list something is "Reserved", we don't have to rely on the list of extensions being the flag. Still, I'm unclear that the core spec. should be including details of an extension before it has been agreed they belong in core.

Agreed. I think it should be easy to filter non-core functionalities when generating the spec because they all have a KHR or company name suffix? How would you think if we generate two html files like the Vulkan spec: one for pure core functionalities, one with all KHR extensions?

johnkslang:
Okay, I suspect this has not been consistently done for all extensions, and indeed makes it quite hard, per-unified, to see what's present and not present.  I will accept this one to increase local consistency, backward compatibly.

The "Reserved." in the unified spec make this work.

> I think it should be easy to filter non-core functionalities when generating the spec because they all have a KHR or company name suffix?

No, I don't think we can rely on spelling.

> How would you think if we generate two html files like the Vulkan spec: one for pure core functionalities, one with all KHR extensions?

For reservations and the connection to capability, I think we are okay now. Are you suggesting full incorporation of extensions?  Or, your main concern is syntax in the grammar file? There is a grey area about how far to go, but, I think we are okay for now not switch to a completely different way of doing extensions.

antiagainst:
> For reservations and the connection to capability, I think we are okay now. Are you suggesting full incorporation of extensions? Or, your main concern is syntax in the grammar file? There is a grey area about how far to go, but, I think we are okay for now not switch to a completely different way of doing extensions.

I'm thinking more of the grammar actually. But discussing the grammar's scope is clearly beyond this pull request. The current status is fine.

PR #61, opened 2018-03-02
Fix Google extension grammar error and backport 

antiagainst:
Fix grammar error for Google decorate_string & hlsl_functionality1.

We don't need to put "LiteralString" as a parameter to OpDecorateString itself;
they should be specified by the Decoration instead.

Note that for the backport, I also included OpDecorateId, but guarded it
with an "extension" field.

Let me know if this is not the preferred way to go. I'd happy to adjust.

### DIFF
--- a/include/spirv/unified1/spirv.core.grammar.json
+++ b/include/spirv/unified1/spirv.core.grammar.json
@@ -3436,9 +3436,7 @@
       "opcode" : 5632,
       "operands" : [
         { "kind" : "IdRef",         "name" : "'Target'" },
-        { "kind" : "Decoration" },
-        { "kind" : "LiteralString" },
-        { "kind" : "LiteralString", "quantifier" : "*" }
+        { "kind" : "Decoration" }
       ],
       "extensions" : [ "SPV_GOOGLE_decorate_string" ],
       "version" : "None"
@@ -3449,9 +3447,7 @@
       "operands" : [
         { "kind" : "IdRef",          "name" : "'Struct Type'" },
         { "kind" : "LiteralInteger", "name" : "'Member'" },
-        { "kind" : "Decoration" },
-        { "kind" : "LiteralString" },
-        { "kind" : "LiteralString",  "quantifier" : "*" }
+        { "kind" : "Decoration" }
       ],
       "extensions" : [ "SPV_GOOGLE_decorate_string" ],
       "version" : "None"
--- a/include/spirv/1.0/spirv.core.grammar.json
+++ b/include/spirv/1.0/spirv.core.grammar.json
@@ -3013,6 +3013,15 @@
       "capabilities" : [ "SparseResidency" ]
     },
     {
+      "opname" : "OpDecorateId",
+      "opcode" : 332,
+      "operands" : [
+        { "kind" : "IdRef", "name" : "'Target'" },
+        { "kind" : "Decoration" }
+      ],
+      "extensions" : [ "SPV_GOOGLE_hlsl_functionality1" ]
+    },
+    {
       "opname" : "OpSubgroupBallotKHR",
       "opcode" : 4421,
       "operands" : [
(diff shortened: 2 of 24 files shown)


include/spirv/unified1/spirv.core.grammar.json
```diff
@@ -3436,9 +3436,7 @@
       "opcode" : 5632,
       "operands" : [
         { "kind" : "IdRef",         "name" : "'Target'" },
-        { "kind" : "Decoration" },
-        { "kind" : "LiteralString" },
-        { "kind" : "LiteralString", "quantifier" : "*" }
+        { "kind" : "Decoration" }
```
dneto0: Right.  In the grammar file the Decoration enum lists its own parameters.

include/spirv/unified1/spirv.core.grammar.json
```diff
@@ -3449,9 +3447,7 @@
       "operands" : [
         { "kind" : "IdRef",          "name" : "'Struct Type'" },
         { "kind" : "LiteralInteger", "name" : "'Member'" },
-        { "kind" : "Decoration" },
-        { "kind" : "LiteralString" },
-        { "kind" : "LiteralString",  "quantifier" : "*" }
+        { "kind" : "Decoration" }
```
dneto0: Agreed, same here.

include/spirv/unified1/spirv.core.grammar.json
antiagainst: Yes. The consumer of the grammar should further consult the grammar of the concrete decoration to determine how to proceed. This is because each decoration can have its own list of parameters, varied of types, we cannot easily represent that here.

johnkslang:
Shouldn't the instruction's specification match the grammar, so that the spec. generation creates the instruction per the specification?

include/spirv/unified1/spirv.core.grammar.json
antiagainst: Sorry. It was my mistake that I forgot about this. `OpDecorate` and friends are all defined likewise.

dneto0:
> Shouldn't the instruction's specification match the grammar, so that the spec. generation creates the instruction per the specification?

The grammar has to be more specific about the exact pattern of arguments for each decoration. E.g. LocalSize has 3 literal integers as arguments. The grammar expresses that on the LocalSize decoration itself.

The spec is more vague in the instruction description of OpDecorate.

johnkslang:
> Note that for the backport, I also included OpDecorateId, but guarded it
with an "extension" field.

Generally, it becomes hard to see what's actually present in the deprecated non-unified spec., especially if the "version" statement is not included.

I think it might be okay to include the "version" statement, to be more clear (does it harm anything?), but preferably, things will be moving to the unified headers.

include/spirv/unified1/spirv.core.grammar.json
johnkslang: I see now, it follows OpDecorate. Thanks. The first PR certainly looked good. :)

antiagainst:
> I think it might be okay to include the "version" statement, to be more clear (does it harm anything?), 

One thing I'm not quite clear about having both "version" and "extensions" is actually how to interpret them together. Because of the current unified1 grammar representation, I think we have an implicit rule requiring them both to be true to enable the opcode? (Extension opcodes have the "version: none" field in unified1 grammar; so I think it means, to use the opcode, the extension should present **and** no version requirement.) 

I assume we want "version : 1.2" here for OpDecorateId. That will turn the requirement into both having the extension and with version >= 1.2?

> but preferably, things will be moving to the unified headers.

Yep. https://github.com/KhronosGroup/SPIRV-Tools/issues/1195 tracks the progress for migrating to unified1 grammar and headers.

dneto0:
> One thing I'm not quite clear about having both "version" and "extensions" is actually how to interpret them together.

> I assume we want "version : 1.2" here for OpDecorateId. That will turn the requirement into both having the extension and with version >= 1.2?

version : "1.2"  would mean it's described in the SPIR-V core spec 1.2 and any later version.  If it's in a core spec then you don't need the OpExtension.

So in a unified grammar having both ' version : "1.2" ' and 'extensions : "blah" '   means that if you're in SPIR-V 1.2 or later you can use without extension.  If you're in SPIR-V 1.0 or 1.1 then you need to declare OpExtension "blah" to use the feature.

johnkslang:
Right. Including the correct "version" makes everything clear. It allows listing extensions without any confusion. It is no longer implicit if something is reserved or in core, it is all explicit, so there is no longer subtle interaction between the fields. Use `"version" : "None"` if something is available only in extensions as of the SPIR-V version stated at the top of the file.

antiagainst:
Cool! That is very clear for me right now. Thanks for the explanation!

PR #62, opened 2018-03-14
Add missing extension/capability requirements for some symbols

antiagainst:
Make the grammar more accurate:

Certain symbols can be enabled by an extension in all SPIR-V
versions, like DeviceGroup. Previously the extension for it
is missing, which makes it look like it is only available from
SPIR-V 1.3.

Make tooling more easier to check certain rules:

Certain symbols require both an extension and a capability to
be available. Instead of only specifying the capability, we can
specify both the capability and the extension on the symbol.
So tools don't need to perform indirect query on the capability
to find out the extension requirements. For example, BaseVertex
requires the DrawParameters capability. But it also requires
the SPV_KHR_shader_draw_parameters extension in SPIR-V <1.3.
Previously we can only find out that fact by checking the
extension requirements for DrawParameters.

Make the grammar more consistent:

Previously some symbols have all their capabilities and extensions
listed, but some are not.

### DIFF
--- a/include/spirv/unified1/spirv.core.grammar.json
+++ b/include/spirv/unified1/spirv.core.grammar.json
@@ -3144,6 +3144,7 @@
         { "kind" : "IdRef", "name" : "'Target'" },
         { "kind" : "Decoration" }
       ],
+      "extensions" : [ "SPV_GOOGLE_hlsl_functionality1" ],
       "version" : "1.2"
     },
     {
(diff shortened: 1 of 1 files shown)


antiagainst:
This pull request is needed to migrate SPIRV-Tools to use unified grammar tables https://github.com/KhronosGroup/SPIRV-Tools/pull/1413.

antiagainst:
Thanks, @johnkslang! :D

PR #63, opened 2018-03-19
Add missing extensions for several symbols

antiagainst:
Added the extension requirement for the symbols of the following
extensions:

* SPV_EXT_shader_stencil_export
* SPV_AMD_shader_explicit_vertex_parameter

This is a follow up of #62.


johnkslang:
LGTM.

There could be no end to missing vendor-specific extensions, including ones you don't even know about. The only requirement is that the token range is reserved (in the registry file) before a vendor starts using it. All tools should work correctly, even in the face of not know the complete set (or any) of the extensions using a token. Always good to collect more information though, and vendors should feel free (encouraged) to make these PRs themselves.

PR #71, opened 2018-04-17
Add a makefile to the top level for convenience

casey:
I found it a bit difficult to figure out how to create the build folder, invoke `cmake`, and correctly invoke the command to rebuild the header files.

This commit adds a makefile to the top level that contains the correct commands, so that a simple `make` in the top level is enough to recompile the header builder and regenerate the headers.


johnkslang:
Oh, this seems to be an OS-specific change. I'm reverting it for now.

PR #79, opened 2018-10-11
Add C# header generation

Thog:
This adds support of C# in buildHeaders.


johnkslang:
Interesting.

Is there a reason to make any headers other than for `unified1`?  I'd recommend only using that.

johnkslang:
Are you sure you want the camel-case spelling in the filename? It doesn't match very well the other spirv.* names.

Thog:
I'm planning on only using ``unified`` but I thought that would be a good idea to have the other versions generated too.
About the camel-case in the file name, I will change it to match the other names as there aren't any conventions about it in C#.

PR #80, opened 2018-10-23
Add support for nv-raytracing-final

alelenv:
@johnkslang for review
Let me know if anything else is needed
SPV Spec
https://gitlab.khronos.org/spirv/spirv-extensions/merge_requests/124

Thanks


johnkslang:
Since some tokens are changing names, existing includers will break and need to change. Is that the intent? Note that we can support aliases, but the order affects the way it will show up in the spec.  Certainly simplest if the clean name change (dropping old token spellings) works.

alelenv:
Yes, we are planning to drop support for nvx version.

PR #83, opened 2018-11-21
Add json support for SPV_EXT_fragment_invocation_density

dgkoch:
- which aliases the SPV_NV_shading_rate functionality


include/spirv/unified1/spirv.core.grammar.json
```diff
@@ -6369,18 +6369,32 @@
           "extensions" : [ "SPV_NV_fragment_shader_barycentric" ],
           "version" : "None"
         },
+        {
+          "enumerant" : "FragSizeEXT",
```
dneto0: Putting the EXT names first will cause the disassembler to prefer the EXT names when outputting.  That's a good choice.  The only impact would be needing an update to SPIRV-Tools tests (if any have been made for the corresponding NV extension.)

include/spirv/unified1/spirv.core.grammar.json
```diff
@@ -7280,7 +7294,14 @@
           "enumerant" : "ShadingRateNV",
           "value" : 5291,
           "capabilities" : [ "Shader" ],
-          "extensions" : [ "SPV_NV_shading_rate" ],
+          "extensions" : [ "SPV_NV_shading_rate", "SPV_EXT_fragment_invocation_density" ],
+          "version" : "None"
+        },
+        {
```
dneto0: I'd put the FragmentDensityEXT spelling first, so swap this element with the previous.  That would match the other enums.

dgkoch:
Updated MR with requested changes.

johnkslang:
Thank you @dneto0. I'll pull this in here and through to the one that generates the spec.

PR #84, opened 2018-11-27
Add WHLSL tool identifier

grorg:


johnkslang:
It seems odd to have "WHLSL" as a vendor. There could be multiple vendors, each making WHLSL tools. Shouldn't the "WHLSL" bit be part of only the "tool" field, and not be a "vendor" field?

Should the vendor be Apple, or WebGPU, or...?

grorg:
OK. Updated the request. @johnkslang - is this ok?

PR #88, opened 2019-01-06
SPV_EXT_physical_storage_buffer

jeffbolznv:


dneto0:
Looks good to me.
I also tested locally with this in SPIRV-Headers and top-of-tree SPIRV-Tools.

PR #94, opened 2019-02-19
Add SPV_NV_cooperative_matrix

jeffbolznv:


include/spirv/unified1/spirv.core.grammar.json
```diff
@@ -4069,6 +4069,75 @@
...
+      "extensions" : [ "SPV_NV_cooperative_matrix" ],
+      "version" : "None"
+    },
+    {
+      "opname" : "OpCooperativeMatrixLengthNV",
+      "opcode" : 5362,
+      "operands" : [
+        { "kind" : "IdResultType" },
+        { "kind" : "IdResult" },
+        { "kind" : "IdRef",        "name" : "'Component Type'" }
```
dneto0: "Component Type" here is confusing/wrong.  The SPIR-V extension calls this "Type".  
See https://github.com/KhronosGroup/SPIRV-Registry/pull/20/files#diff-853100fe9e5b756198d2fc810a42ab8dR290

dneto0:
Added @johnkslang as reviewer.  He does the merges on this repo.

include/spirv/unified1/spirv.core.grammar.json
jeffbolznv: Good catch, this was a copy/paste bug from OpTypeCooperativeMatrixNV. Fixed.

include/spirv/unified1/spirv.core.grammar.json
```diff
@@ -4069,6 +4069,75 @@
...
+      "extensions" : [ "SPV_NV_cooperative_matrix" ],
+      "version" : "None"
+    },
+    {
+      "opname" : "OpCooperativeMatrixMulAddNV",
+      "opcode" : 5361,
+      "operands" : [
+        { "kind" : "IdResultType" },
+        { "kind" : "IdResult" },
+        { "kind" : "IdRef",             "name" : "'A'" },
```
johnkslang: These operand names stand out as being different (unusually short).  Typical arithmetic ops are like *Operand 1*, but a more descriptive name would be even better, given the three operands.

At reservation time, it is your choice.

jeffbolznv: A/B/C are pretty common names for matrix multiply-add (A*B+C), and I think the SPIR-V instruction description reads better with these as A/B/C than they would with "Operand 1/2/3". So my preference is to leave them as they are.

dneto0:
I have verified this grammar change with SPIRV-Tools top of tree and  with the cooperative-matrix MR for SPIRV-Tools (https://github.com/KhronosGroup/SPIRV-Tools/pull/2404 ).  That other MR has its own issues, but they are not caused by this grammar change.

PR #100, opened 2019-03-29
Add a function that returns whether an opcode has a result and/or result type

jeffbolznv:
For code that needs to do minor inspection/validation of SPIR-V it is often useful to be able to build an id->typeid map, but there's currently no good way to know which opcodes have a result/type and which don't, and this makes it clumsy to try to build the map.

My original motivation for this is at: https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/742/files#diff-02aa55cdf8efb86e40256744eb74b841R1949. I needed to do some validation of cooperative matrix types, but the cooperative matrix operands could be a result of a bunch of different instructions. There were few enough that it was reasonable to whitelist the instructions in this case.

Baldur described a case where he needed to know what type of integer an OpAccessChain operand is. This would be easy to know by building an id->typeid map and only inspecting OpTypeInt and OpAccessChain instructions.

So I think this is generally useful, and the JSON has this information encoded in it. The core SPIR-V header seems like a natural place to put this, though I can imagine somebody objecting to having utility code creep into the header.


jeffbolznv:
@johnkslang @dneto0 @alan-baker I'd like to get your opinions on this.

dj2:
What about putting it into spirv-tools instead of the header?

jeffbolznv:
Just as a practical matter, SPIRV-Tools doesn't check in code generated from a particular commit of the JSON grammar, it's always generated at build time from whatever JSON is fetched. So if it were part of SPIRV-tools, using it would require building it from source and that probably makes it not palatable for most people.

Also, SPIRV-Tools already has this information in it (see spv_opcode_desc_t and core.insts-unified1.inc) so it would be redundant. Maybe it could make sense to try to factor that out into something that is generated in the SPIRV-Headers repo, but even that seems may be impractical.

I think the target users of this want something really simple to incorporate into their code with as little overhead/effort as possible, and always want it to stay up to date with the latest grammar changes. So having it directly in SPIRV-Headers seems very natural to me.

johnkslang:
I'd rather not see it, but if it is there, wouldn't it be better tabled instead of being procedural?

jeffbolznv:
The opcode values aren't tightly packed, so a table is pretty wasteful.

johnkslang:
It depends on the table, and actually the JSON grammar is the table format for all of this, so I realize now the right way to do this is to add these to the instructions in the JSON grammar, which is why something didn't feel right about this.

I also think op-code class should get added to the JSON, as it's typically not dealt with in reservations, causing them to show up in the wrong place in the spec until someone manually fixes them.

I'm willing to add the type/result/op-class if others agree.  Should be backward compatible to add fields.  @jeffbolznv, @dneto0 ?

johnkslang:
Actually, the return and result are already in the JSON grammar, for opcodes that have them:

```
        { "kind" : "IdResultType" },
        { "kind" : "IdResult" },
```

These are the tabled form that the PR (indirectly) uses to create the procedural form in the headers.

jeffbolznv:
Right, the goal is to have this in a C/C++ header (and the other generated headers, too, eventually).

I'm fine with trying to find a more complete way to expose all the information from the JSON. This was just solving the need we've actually seen in a few tools.

dneto0:
I agree it's backward compatible to add more fields to the JSON data; this would be for the instruction class, to group them in the spec.

I support adding the functions directly to the headers, provided they only add to code size when they are actually referenced.  I think making them inline is sufficient.  If there are concerns then we could guard them with a preprocessor symbol or put them in a different header file.

Yes, the token numbers are somewhat sparse.  I'm not too concerned provided you only pay the cost if you use the functions.
The SPIRV-Tools tables are sorted by token number and we use a binary search. (Even so they are somewhat wasteful of space and there are relatively simple ways to make it more dense, but that's future work.)

jeffbolznv:
As we discussed in today's meeting, I added an #ifdef around the new utility code. I've also test compiled all the C/C++/C++11 headers and fixed some build issues in them.

johnkslang:
It looks good to me.

Ping me when de-WIP.

jeffbolznv:
OK, removed WIP. @johnkslang ready to merge.

PR #103, opened 2019-05-09
Enable Groups capability by the extension SPV_AMD_shader_ballot

amdrexu:
This is to address the issue: https://github.com/KhronosGroup/glslang/issues/1763.


dneto0:
This looks good to me.  It resolves https://github.com/KhronosGroup/glslang/issues/1763

PR #116, opened 2019-06-03
Add SPV_EXT_fragment_shader_interlock and SPV_NV_shader_sm_builtins

jeffbolznv:


johnkslang:
I think we just had some bad timing here, given that I already merged one of these? I assume now, we should just keep the other individual one.

jeffbolznv:
Yeah, I just rebased 114. Closing this one.

PR #117, opened 2019-06-10
Add grammar and symbols for UserTypeGOOGLE extension to unified1.

ehsannas:


include/spirv/unified1/spirv.py
```diff
@@ -428,6 +428,7 @@
         'HlslCounterBufferGOOGLE' : 5634,
         'HlslSemanticGOOGLE' : 5635,
         'UserSemantic' : 5635,
+        'UserTypeGOOGLE' : 5635,
```
dneto0: This is using the wrong number.  But I'm about to submit a PR that fixes this (a side effect of using the header generation tool)

ehsannas: Thanks

PR #119, opened 2019-06-12
Add Volatile to Memory Semantics, for SPV_KHR_vulkan_memory_model

dneto0:
From Khronos-internal issue 476: Update SPV_KHR_vulkan_memory_model
to be able to express volatile OpAtomic* operations.

Also regenerate the language-specific headers.  This step fixes
two problems with UserTypeGOOGLE in those headers.

### DIFF
--- a/include/spirv/unified1/spirv.core.grammar.json
+++ b/include/spirv/unified1/spirv.core.grammar.json
@@ -6120,6 +6120,12 @@
           "value" : "0x4000",
           "capabilities" : [ "VulkanMemoryModelKHR" ],
           "version" : "None"
+        },
+        {
+          "enumerant" : "Volatile",
+          "value" : "0x8000",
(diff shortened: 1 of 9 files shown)


dneto0:
Cc: @alan-baker

include/spirv/unified1/spirv.core.grammar.json
```diff
@@ -6120,6 +6120,12 @@
           "value" : "0x4000",
           "capabilities" : [ "VulkanMemoryModelKHR" ],
           "version" : "None"
+        },
+        {
+          "enumerant" : "Volatile",
```
jeffbolznv: @dneto0 I missed this change yesterday. This should have a KHR suffix.

johnkslang: Because Khronos already agreed to final semantics for this spelling, it is okay to not use the suffix; we get to skip the aliasing process in this situation.

PR #121, opened 2019-07-01
add SPV_EXT_demote_to_helper_invocation

jeffbolznv:


johnkslang:
Order ends up mattering (combination of parallel development and git), and I keep fixing it, but not always noticing how it gets in.

Could you restore this and Nvidia's other recent addition to numerical order?

Thanks.

jeffbolznv:
Fixed the order.

PR #132, opened 2019-10-10
Skip spirv headers installation

lukaszgotszaldintel:
Move not necessary instructions when skip SPIRV-Headers installation.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -68,16 +68,16 @@ if (SPIRV_HEADERS_ENABLE_EXAMPLES)
   add_subdirectory(example)
 endif()
 
-include(GNUInstallDirs)
-add_library(${PROJECT_NAME} INTERFACE)
-target_include_directories(${PROJECT_NAME} INTERFACE
-    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
-)
-
 # Installation
 
 if (SPIRV_HEADERS_ENABLE_INSTALL)
     message(STATUS "Installing SPIRV-Header")
+
+    include(GNUInstallDirs)
+    add_library(${PROJECT_NAME} INTERFACE)
+    target_include_directories(${PROJECT_NAME} INTERFACE
+        $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+    )
 
     set(config_install_dir "${CMAKE_INSTALL_LIBDIR}/cmake/${PROJECT_NAME}")
 


lukaszgotszaldintel:
These instructions are not needed if you are skipping the SPIRV-Headers installation. I have found a problems with them under Visual Studio when compiling together with SPIRV-Tools:

CMake Error in SPIRV-Headers/CMakeLists.txt:
  INTERFACE_LIBRARY targets may only have whitelisted properties.  The
  property "FOLDER" is not allowed.

dj2:
Doesn't this now mean that SPIRV-Headers no longer exists as a project which can be used by others? The add_library change seems unrelated to the error you're getting, same with the install line?

What line was producing the error you listed above? I don't see the property `FOLDER` added? or is that done by the `target_include_directories`?

lukaszgotszaldintel:
The lines I have moved have no meaning when skipping SPIRV-Headers installation. In this case user will only do add_subdirectory for SPIRV-Headers directory and all is fine.

I have found that error is related only to "INTERFACE" and "FOLDER". The second one is used in SPIRV-Tools project many times:

set_property(TARGET ${ARG_TARGET} PROPERTY FOLDER "SPIRV-Tools executables")
set_property(TARGET ${target} PROPERTY FOLDER "SPIRV-Tools tests")
set_property(TARGET SPIRV-Tools-reduce PROPERTY FOLDER "SPIRV-Tools libraries")
set_property(TARGET SPIRV-Tools-reduce PROPERTY FOLDER "SPIRV-Tools libraries")
set_property(TARGET SPIRV-Tools-opt PROPERTY FOLDER "SPIRV-Tools libraries")
set_property(TARGET SPIRV-Tools-link PROPERTY FOLDER "SPIRV-Tools libraries")
set_property(TARGET SPIRV-Tools-fuzz PROPERTY FOLDER "SPIRV-Tools libraries")
set_property(TARGET ${ARG_TARGET} PROPERTY FOLDER "SPIRV-Tools examples")
set_property(TARGET ${SPIRV_TOOLS} PROPERTY FOLDER "SPIRV-Tools libraries")
set_property(TARGET ${SPIRV_TOOLS}-shared PROPERTY FOLDER "SPIRV-Tools libraries")
set_property(TARGET spirv-tools-build-version PROPERTY FOLDER "SPIRV-Tools build")
set_property(TARGET spirv-tools-vimsyntax PROPERTY FOLDER "SPIRV-Tools utilities")
set_property(TARGET spirv-tools-header-${NAME} PROPERTY FOLDER "SPIRV-Tools build")
set_property(TARGET spv-tools-${SHORT_NAME} PROPERTY FOLDER "SPIRV-Tools build")

Have you tried to move the lines and skip installation? What do you think, please?

dj2:
I haven't, just trying to understand what is causing the need for the change.

lukaszgotszaldintel:
I see. Well, the need of this change is to compile SPIRV-Headers project and SPIRV-Tools project together in a single Visual Studio solution. How is that possible since first is "INTERFACE" and second use property "FOLDER", please? "INTERFACE" has been added on 2019-05-11 by mchiasson (commit c5a291f7a5c35a878c4a1cd37db7b85eef1c6f71). So far I was using earlier version of both projects and all was good.

johnkslang:
@dj2 are you okay with this change now?

dj2:
Sorry about the slow reply. I guess my confusion comes from, I don't see `FOLDER` being used here. So, is the issue how SPIRV-Tools uses SPIRV-Headers? Is the correct change in the tools repo? Adding the include directories seems like it shouldn't be contingent on using INSTALL?

lukaszgotszaldintel:
Yes, you are right, it is all about how SPIRV-Tools uses SPIRV-Headers. Finally I have decided to remove the "add_subdirectory" for SPIRV-Headers from our project and ignore it's CMakeLists.txt. We will only use "include_directories" as workaround.

https://github.com/intel/intel-graphics-compiler/commit/1f5193eac83392fd964113364fc5303b7548e7c4

You can ignore this pull request, please.

PR #135, opened 2020-01-01
Also propagate SPIRV-Headers version to CMakeLists.txt

vulturm:
Was missed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,7 +29,7 @@
 # https://www.khronos.org/registry/spir-v/
 #
 cmake_minimum_required(VERSION 3.0)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -29,7 +29,7 @@
 # https://www.khronos.org/registry/spir-v/
 #
 cmake_minimum_required(VERSION 3.0)
-project(SPIRV-Headers VERSION 1.4.1)
+project(SPIRV-Headers VERSION 1.5.1)
```
johnkslang: What does this get used for?  I'd rather eliminate it than have yet another independent version number that must be kept in sync.

PR #136, opened 2020-01-04
Fix Vulkan memory model's extension specification

antiagainst:
VulkanKHR becomes Vulkan since SPIR-V 1.5; so Vulkan is just a symbol
duplicate of VulkanKHR. It should have all properties like VulkanKHR,
including the enabling extension.

### DIFF
--- a/include/spirv/unified1/spirv.core.grammar.json
+++ b/include/spirv/unified1/spirv.core.grammar.json
@@ -7118,6 +7118,7 @@
           "enumerant" : "Vulkan",
           "value" : 3,
           "capabilities" : [ "VulkanMemoryModel" ],
+          "extensions" : [ "SPV_KHR_vulkan_memory_model" ],
           "version" : "1.5"
         },
         {


include/spirv/unified1/spirv.core.grammar.json
```diff
@@ -7118,6 +7118,7 @@
           "enumerant" : "Vulkan",
           "value" : 3,
           "capabilities" : [ "VulkanMemoryModel" ],
+          "extensions" : [ "SPV_KHR_vulkan_memory_model" ],
```
johnkslang: The `VulkanKHR` alias for this, a few lines below, does have this listed.  If the argument is these are aliases and hence should say exactly the same thing, that really indicates we should just be listing aliases as an entry here, not having complete duplicated entries, and we've previously agreed that's what we want to do.

The only real reason to keep separate entries is if they don't express identical things, and it is pretty sensible to list the extension for the extension spelling, while not listing it for the core spelling.  That goes hand-in-hand with listing the core semantics with the core spelling and referring to the extension for the extension semantics.

Well, both views make sense:
- two separate entries that show differences
- one entry that lists aliases

I recommend we finalize on one of these and start doing that.

antiagainst: Some background on why I proposed this fix: I was generating op availability information for the SPIR-V dialect using the grammar and tried to use the canonical symbol for various enumerants. So instead of pulling the infromation from `VulkanKHR`, my [code](https://github.com/antiagainst/llvm-project/commit/66d36cfa54f75206acbb8bc980cbe700d6131ce5#diff-34d33579b227846005e247d0bd539b8a) was using `Vulkan`, and then I found that the extension is missing. IIUC, `Vulkan` is just a canonical symbol for `VulkanKHR` given they map to the same value. If I'm targeting SPIR-V pre-1.5 but still have the Vulkan memory model extension then I should still be fine to use the `Vulkan` symbol (which has the same value as `VulkanKHR`); but that is missing in the spec as the current state.

Agreed that we should have a consistent rule for this! As a comparison we have `PhysicalStorageBuffer64EXT` and `PhysicalStorageBuffer64`, while `PhysicalStorageBuffer64` lists more complicated enabling extensions than even `PhysicalStorageBuffer64EXT`. IMO, this inconsistency is confusing at least. From a tooling's perspective, I vote for 2) one entry that lists aliases. I think it has the following benefits:

* It simplifies the grammar file. I'd imagine in the future we will have more and more extension symbol vs core symbol cases. Manually duplicating them and keeping consistency is burdensome.
* It simplifies the downstream tooling. Now tools don't need to be clever regarding symbol suffix and "guess" which is the canonical one and figure out the differences.
* It shows the relationship between these symbols better.

As for the detailed grammar, I agree that we can probably add a `aliases` field.

antiagainst: Any advise on how to proceed?

antiagainst: Should I propose a new PR to implement the alias method as discussed above?

johnkslang: I think we should all invest in the alias method, not other paths.  Tooling needs to be involved.  I guess we could start by adding alias data to the existing header, as adding things shouldn't break anything.  Looking at how detailed the steps could be:

. alias field for just what you care about, nothing else added or removed
. alias field for all things that are aliased
. SPIR-V specification generation tooling updated to understand it
. SPIRV-Tools and other things(?) updated to understand it
. redundant existing way of duplicating entries is eliminated

If you're really up for the first 1 or 2 of those, I think that would be fine.

antiagainst: Sure, I'll take a stab at adding the aliases without deleting existing duplicated ones. Thanks John!

antiagainst: Okay, I've created #138 to introduce `aliases` and #139 to remove duplicated cases. PTAL!

antiagainst:
Closing this one in favor of #138 and #139.

PR #139, opened 2020-01-19
Remove duplicated cases in favor of 'aliases' fields

antiagainst:
This is built on top of #138: the first commit is from #138, so only the second one is interesting here.


Naghasan:
closing in favour of https://github.com/KhronosGroup/SPIRV-Headers/pull/447

PR #140, opened 2020-01-20
Add BUILD.gn

ShabbyX:
Allows gn builds to include files from spirv-headers.

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -1,0 +1,32 @@
+# Copyright 2020 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


BUILD.gn
```diff
@@ -0,0 +1,32 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+config("spv_headers_public_config") {
+  include_dirs = [ "include" ]
+}
+
+source_set("spv_headers") {
+  sources = [
+    "include/spirv/1.2/GLSL.std.450.h",
```
dj2: I don't think you need the 1.2 as they should all be overridden by the unified1 entries.

@dneto0  is that correct?

ShabbyX: I see one of the sources in the BUILD.gn of spirv_tools references 1.2 instead of unified1 (`GLSL.std.450.h`). I have a feeling it might have to do with some project that includes `1.2/GLSL.std.450.h` instead of `unified1/GLSL.std.450.h`.

I was thinking of keeping it so spirv-tools/BUILD.gn can switch to using this right away. I don't know which user of GLSL.std.450.h is including 1.2/ to fix that (if any). According to cs.chromium.org, it's not Chromium.

dj2: Fair enough, could be angle or some other downstream consumer.

PR #143, opened 2020-02-21
Add grammars, C header, and header generator for vendor and KHR extended instruction sets

dneto0:


dneto0:
cc: @jeffbolznv

dneto0:
@jaebaek and @ben-clayton were also interested in this.

include/spirv/unified1/AMD_shader_epxlicit_vertex_parameter.h
```diff
@@ -0,0 +1,47 @@
...
+// 
+
+#ifndef SPIRV_UNIFIED1_AMD_shader_epxlicit_vertex_parameter_H_
+#define SPIRV_UNIFIED1_AMD_shader_epxlicit_vertex_parameter_H_
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+enum { AMD_shader_epxlicit_vertex_parameterRevision = 4, AMD_shader_epxlicit_vertex_parameterRevision_BitWidthPadding = 0x7fffffff };
```
johnkslang: This enum is missing newlines and does not match the other enums, should they match?

tools/buildHeaders/bin/makeExtinstHeaders
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
```
johnkslang: I've gotten some push back on bash being platform dependent.  Is there a reason to use bash instead of python?

tools/buildHeaders/bin/makeHeaders
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+sh bin/makeExtinstHeaders
```
johnkslang: Well, this bash partially answers the previous question.  Let's say my question is still valid, would like an answer, but not necessarily asking for a change, just the answer.

include/spirv/unified1/AMD_shader_epxlicit_vertex_parameter.h
dneto0: Hmm. Yes, it should match.
I also note the header name is spelled wrong. (!)

tools/buildHeaders/bin/makeExtinstHeaders
dneto0: I think I was following the existing script example.  I can rewrite in Python3.
(and if we go that route, we can eliminate the need for dos2unix)

dneto0: Rewritten as Python3

include/spirv/unified1/AMD_shader_epxlicit_vertex_parameter.h
dneto0: Fixed both the names, and inserted newlines.

tools/buildHeaders/bin/makeHeaders
dneto0: Fixed

dneto0:
All feedback addressed. PTAL

PR #149, opened 2020-03-17
Add SPV_KHR_ray_{tracing,query} to headers

dgkoch:
cc @johnkslang


johnkslang:
I have confirmed that a local build of this merge replicates the same header contents, when regenerating headers.

PR #154, opened 2020-04-18
OpCopyMemory and OpCopyMemorySized grammar type fixed.

gleblebedev:
It seems that OpCopyMemory and OpCopyMemorySized has duplicated MemoryAccess access operand.


johnkslang:
I don't think this needs a fix:

image

gleblebedev:
Sorry, was looking at outdated documentation

PR #163, opened 2020-07-20
Support SPV_KHR_terminate_invocation

alan-baker:


include/spirv/unified1/spirv.core.grammar.json
```diff
@@ -4127,6 +4127,16 @@
       "extensions" : [ "SPV_KHR_shader_ballot" ],
       "version" : "None"
     },
+    {
```
johnkslang, line 4130: These need to stay in numerical order.

PR #164, opened 2020-07-27
Publish the headers for the clspv embedded reflection non-semantic extended instruction set

alan-baker:
Headers for https://github.com/KhronosGroup/SPIRV-Registry/pull/78


include/spirv/unified1/extinst.nonsemantic.clspvreflection.grammar.json
```diff
@@ -0,0 +1,237 @@
...
+        { "kind" : "IdRef", "name" : "Kernel" },
+        { "kind" : "IdRef", "name" : "Name" }
+      ]
+    },
+    {
+      "opname" : "ArgumentInfo",
+      "opcode" : 2,
+      "operands" : [
+        { "kind" : "IdRef", "name" : "Name" },
+        { "kind" : "IdRef", "name" : "Type Name", "quantifier" : "?" },
```
dneto0, line 17: I don't recall seeing multiple optional operands in sequence.
Does the SPIRV-Tools parser handle this? I assume that operands are consumed left to right, so if I only had two operands then it would be TypeName and AddressQualifier.

But if this works, then great.

alan-baker: Yes, the parser handles this. Your assumption is correct about multiple optional operands. We already have an instance of this for OpCopyMemory.

PR #166, opened 2020-07-29
Update BUILD.bazel and BUILD.gn

alan-baker:
* Export NonSemantic.ClspvReflection.h for both
* Add exports for the extended instruction sets in the unified1
  directory (for use in SPIRV-Tools)

This is necessary for https://github.com/KhronosGroup/SPIRV-Tools/pull/3618

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -71,6 +71,46 @@ filegroup(
     srcs = ["include/spirv/spir-v.xml"],
 )
 
+filegroup(
+    name = "spirv_ext_inst_debuginfo_grammar_unified1",
+    srcs = ["include/spirv/unified1/extinst.debuginfo.grammar.json"],
+)
+
+filegroup(
+    name = "spirv_ext_inst_nonsemantic_clspvreflection_grammar_unified1",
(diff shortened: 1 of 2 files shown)


BUILD.bazel
```diff
@@ -71,6 +71,46 @@ filegroup(
...
+filegroup(
+    name = "spirv_ext_inst_spv_amd_shader_explicit_vertex_parameter_grammar_unified1",
+    srcs = ["include/spirv/unified1/extinst.spv-amd-shader-explicit-vertex-parameter.grammar.json"],
+)
+
+filegroup(
+    name = "spirv_ext_inst_spv_amd_shader_trinary_minmax_grammar_unified1",
+    srcs = ["include/spirv/unified1/extinst.spv-amd-shader-trinary-minmax.grammar.json"],
+)
+
```
dnovillo, line 113: These filegroups() don't seem to be used?

dnovillo: Saw the uses in https://github.com/KhronosGroup/SPIRV-Tools/pull/3618.  Another alternative you may want to use is to declare these as `export_files` (https://docs.bazel.build/versions/master/be/functions.html#exports_files) which explicitly states that these files are to be used in other packages.

PR #171, opened 2020-10-10
Register the Xenia emulator as a generator

Triang3l:
An inconvenient way to be honest, would be much better if this was a string in shader modules (like in DXBC, which isn't even really supposed to be emitted by anything but Microsoft's tools)… [Xenia](https://github.com/xenia-project/xenia) is an Xbox 360 emulator capable of running hundreds of commercial games, we are currently in the process of rewriting its Vulkan rendering backend, which includes a translator from ATI Xenos microcode to SPIR-V, plus emission of emulation-specific SPIR-V code (such as fragment shader interlock-based programmable pixel packing/blending). No references to the Xbox 360 or the Xenos made in the file or the commit message, so no potential trademark violation is done.

### DIFF
--- a/include/spirv/spir-v.xml
+++ b/include/spirv/spir-v.xml
@@ -76,7 +76,8 @@
         <id value="23"  vendor="Google" tool="Tint Compiler" comment="Contact David Neto, dneto@google.com"/>
         <id value="24"  vendor="Google" tool="ANGLE Shader Compiler" comment="Contact Shahbaz Youssefi, syoussefi@google.com"/>
         <id value="25"  vendor="Netease Games" tool="Messiah Shader Compiler" comment="Contact Yuwen Wu, atyuwen@gmail.com"/>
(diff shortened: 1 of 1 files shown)


Triang3l:
Should I also register a SourceLanguage (I don't really want to do that because it's versioned, and I'm planning to require the minimum of 1.0), or is Unknown fine too? Not sure what to call it — it is a variant of GL_AMD_shader_binary_Z400, but GL_AMD_shader_binary_Z400 is used even for very different binaries for Adreno 3xx or later. Among ones currently somewhat used in not completely retrogaming contexts, could be SourceLanguageZ430, but has minor differences in texture fetch instructions from AMD Z430 (Adreno 200). SourceLanguageXenos? Though SourceLanguageUnknown seems like the one with the least questions.

johnkslang:
You are not required to register the source language.

PR #184, opened 2020-11-27
remove HitTKHR alias

dgkoch:
It was not added to the SPV_KHR_ray_tracing extension since it is just
an alias of RayTMaxKHR.

### DIFF
--- a/include/spirv/unified1/spirv.core.grammar.json
+++ b/include/spirv/unified1/spirv.core.grammar.json
@@ -10431,15 +10431,8 @@
         {
           "enumerant" : "HitTNV",
           "value" : 5332,
-          "capabilities" : [ "RayTracingNV" , "RayTracingKHR" ],
-          "extensions" : [ "SPV_NV_ray_tracing" , "SPV_KHR_ray_tracing" ],
-          "version" : "None"
-        },
-        {
-          "enumerant" : "HitTKHR",
-          "value" : 5332,
-          "capabilities" : [ "RayTracingNV" , "RayTracingKHR" ],
-          "extensions" : [ "SPV_NV_ray_tracing" , "SPV_KHR_ray_tracing" ],
+          "capabilities" : [ "RayTracingNV" ],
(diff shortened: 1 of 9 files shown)


dgkoch:
cc @raunraun @johnkslang

raunraun:
I'm away from a computer to do a test build.  Visually looks good.  I'm going to merge and fix later tonight if needed.

dgkoch:
Thanks!

dgkoch:
Pulling this into spirv-tools here: https://github.com/KhronosGroup/SPIRV-Tools/pull/4040 

(I'm writing some spirv-tools tests that will depend on it not being supported)

dgkoch:
Glslang fixes https://github.com/KhronosGroup/glslang/pull/2471

johnkslang:
Note this was a backward incompatible change to the header.  That should almost never happen, and doesn't need to happen just to remove an alias does it?

Anyone who got a header and then used that token will now break.

johnkslang:
When we're confident with this fix we should also re tag, so people are getting the fix instead of the current one tagged for ray tracing (1.5.4.raytracing).

dgkoch:
> That should almost never happen, and doesn't need to happen just to remove an alias does it?

Is there another way to remove the alias that wasn't backwards incompatible?

And yes I'm confident in this fix, or I wouldn't have proposed it.

johnkslang:
> And yes I'm confident in this fix, or I wouldn't have proposed it.

Sorry, I didn't say that right.  I meant more something like are there any other fixes possibly coming?  As in, tag now with a 1.5.4.raytracing.fixed or wait a bit?

But, it does seem like we'd want to stop indicating to grab to the 1.5.4.raytracing tag and get something more recent.

dgkoch:
It would be nice to get https://github.com/KhronosGroup/SPIRV-Tools/pull/4041 in yet, but even without that, yes I agree we should have a new tag.

PR #187, opened 2021-01-05
add function control bitfield reservation section

bashbaug:
Add a section to the XML file for function control bitfield reservations and reserve bit 16 for an upcoming Intel extension.

This new section uses the same layout and strategy as the existing sections for loop control and fp fast math mode bitfield reservations.


mkinsner:
Looks good to me.

PR #190, opened 2021-01-20
Header generator: Check enumerant ordering

dneto0:
In the grammar, enforce ordering rules:
- Instructions must appear in order of their opcode
- Non-instructions: each successive enumerant within a single kind must
  have a value which is either larger than all the previous values
  or equal to one of the previous values.
  We have this convoluted rule for enums because we want to support
  the sequence:
  	4416 SubgroupEqMask
  	4417 SubgroupGeMask
  	4418 SubgroupGtMask
  	4419 SubgroupLeMask
  	4420 SubgroupLtMask
  	4416 SubgroupEqMaskKHR
  	4417 SubgroupGeMaskKHR
  	4418 SubgroupGtMaskKHR
  	4419 SubgroupLeMaskKHR
  	4420 SubgroupLtMaskKHR

### DIFF
--- a/tools/buildHeaders/jsonToSpirv.cpp
+++ b/tools/buildHeaders/jsonToSpirv.cpp
@@ -25,6 +25,7 @@
 #include <assert.h>
 #include <string.h>
 #include <algorithm>
+#include <cstdlib>
 #include <iostream>
 #include <unordered_map>
 #include <unordered_set>
@@ -326,6 +327,8 @@ void jsonToSpirv(const std::string& jsonPath, bool buildingHeaders)
 
     // process the instructions
     const Json::Value insts = root["instructions"];
+    unsigned maxOpcode = 0;
+    bool firstOpcode = true;
     for (const auto& inst : insts) {
         const auto printingClass = inst["class"].asString();
         if (printingClass.size() == 0) {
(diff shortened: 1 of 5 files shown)


tools/buildHeaders/jsonToSpirv.cpp
```diff
@@ -335,6 +338,19 @@ void jsonToSpirv(const std::string& jsonPath, bool buildingHeaders)
...
         const std::string name = inst["opname"].asString();
+        if (firstOpcode) {
+          maxOpcode = opcode;
+          firstOpcode = false;
+        } else {
+          if (maxOpcode > opcode) {
+            std::cerr << "Error: " << name
+                      << " is out of order. It follows the instruction with opcode " << maxOpcode
+                      << std::endl;
+            std::exit(1);
```
dneto0, line 349: This may be controversial.  There are also other errors which do not cause immediate termination.
However, those errors are not propagated out to the exit code of the program. I think they should, but can do that separately.

dneto0:
I got feedback that enumerants should appear in order (i.e. use the simpler rule), and to adjust the ordering in the JSON grammar file to satisfy the rule.

I've force-pushed the change to do that.

PR #193, opened 2021-01-27
add None as a possible value for DebugInfoFlags

bashbaug:
It's possible that no debug info flags are needed, in which case the debug info flags will have value `0`.  This case should be disassembled into `None` and then `None` should be assembled back into `0`.  Without this change many SPIR-V modules with debug information cannot be properly disassembled and assembled.

I think this may require a minor spec update as well?  I'll work on that next.

CC: @AlexeySotkin

### DIFF
--- a/include/spirv/unified1/DebugInfo.h
+++ b/include/spirv/unified1/DebugInfo.h
@@ -78,6 +78,7 @@ enum DebugInfoInstructions {
 
 
 enum DebugInfoDebugInfoFlags {
+    DebugInfoNone = 0x0000,
     DebugInfoFlagIsProtected = 0x01,
     DebugInfoFlagIsPrivate = 0x02,
     DebugInfoFlagIsPublic = 0x03,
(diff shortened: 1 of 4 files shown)


bashbaug:
Hmm, I may need to regenerate the headers for these extended instruction sets also?  I'll do this shortly.

AlexeySotkin (approved):
LGTM, thanks!

bashbaug:
Ping @dneto0 @johnkslang @raunraun - could one of you please review and merge this PR assuming everything looks good?  It's holding up some CTS tests.  Thanks!

iamen12345:
> It's possible that no debug info flags are needed, in which case the debug info flags will have value `0`. This case should be disassembled into `None` and then `None` should be assembled back into `0`. Without this change many SPIR-V modules with debug information cannot be properly disassembled and assembled.
> 
> I think this may require a minor spec update as well? I'll work on that next.
> 
> CC: @AlexeySotkin

hzjxnxjx

PR #197, opened 2021-02-19
Reserve loop control bit 24 for placeholder/passthrough controls

mkinsner:
Needed for upcoming Intel loop control extension


mkinsner:
Closing PR temporarily while we confirm that we can't leverage an existing bit allocation.

PR #210, opened 2021-05-30
Add SpvGenTwo tools to vendor IDs

rAzoR8:
SpvGenTwo https://github.com/rAzoR8/SpvGenTwo is a SPIR-V IR Builder / Parser toolbox with disassembler, reflector, linker and more.


rAzoR8:
Sorry for the faulty entry and thanks for fixing!

PR #211, opened 2021-06-02
Fix OpExecutionMode json specification

ShabbyX:
This instruction takes a list of literals which was missing from the
json specification.

Signed-off-by: Shahbaz Youssefi <ShabbyX@gmail.com>


dneto0 (commented):
This isn't how the grammar works.
The literal operands, if they exist, are associated with the enum value itself.
E.g. see LocalSize has 3 literal integer parameters.

The JSON grammar is more precise than the way it's presented in the instruction description in the SPIR-V spec.

dneto0:
This would actually break tooling.

PR #215, opened 2021-06-08
buildHeaders: Add override decoration

dneto0:
Fixes warnings in AppleClang

### DIFF
--- a/tools/buildHeaders/header.cpp
+++ b/tools/buildHeaders/header.cpp
@@ -500,7 +500,7 @@ namespace {
 
         virtual std::string fmtEnumUse(const std::string& opPrefix, const std::string& name) const { return pre() + name; }
 
-        virtual void printHasResultType(std::ostream& out) const
+        virtual void printHasResultType(std::ostream& out) const override
         {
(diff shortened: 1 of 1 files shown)


alan-baker (approved):

dneto0:
Without this, I get a warning in AppleClang

e.g.

../header.cpp:646:29: warning: 'fmtEnumUse' overrides a member function but is not marked 'override' [-Winconsistent-missing-override]
        virtual std::string fmtEnumUse(const std::string& opPrefix, const std::string& name) const { return opPrefix + "::" + name; }
                            ^
../header.cpp:501:29: note: overridden virtual function is here
        virtual std::string fmtEnumUse(const std::string& opPrefix, const std::string& name) const { return pre() + name; }
                            ^

The override specifier is new in C++11.  The rest of the code already assumes C++11, and the CMake file requests it for GCC and Clang like compilers.   I don't have an MSVC build handy, and there's no CI on this. In any case, this subproject is only used by folks who update the generated headers, and is not (to my knowledge) used by projects that incorporate SPIRV-Headers as a subproject.

PR #221, opened 2021-06-16
Remove the need to install buildSpvHeaders

kpet:
The README file currently doesn't mention the need to call the install
target. Use the built binary directly.

If requiring installation makes sense for other reasons, I'm happy to edit the
the README instead.

Signed-off-by: Kevin Petit <kevin.petit@arm.com>

### DIFF
--- a/tools/buildHeaders/bin/makeHeaders
+++ b/tools/buildHeaders/bin/makeHeaders
@@ -3,5 +3,5 @@
 python3 bin/makeExtinstHeaders.py
 
 cd ../../include/spirv/unified1
-../../../tools/buildHeaders/build/install/bin/buildSpvHeaders -H spirv.core.grammar.json
+../../../tools/buildHeaders/build/buildSpvHeaders -H spirv.core.grammar.json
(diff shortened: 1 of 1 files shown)


tools/buildHeaders/bin/makeHeaders
```diff
@@ -3,5 +3,5 @@
 python3 bin/makeExtinstHeaders.py
 
 cd ../../include/spirv/unified1
-../../../tools/buildHeaders/build/install/bin/buildSpvHeaders -H spirv.core.grammar.json
+../../../tools/buildHeaders/build/buildSpvHeaders -H spirv.core.grammar.json
```
dneto0, line 6: Is this valid for Windows builds?  In that case it might be in a "Release" directory, or whatever the configuration name is.

kpet: Not sure, I didn't test Windows. I don't want to break anybody's workflow so I'll close this PR and suggest https://github.com/KhronosGroup/SPIRV-Headers/pull/222 instead (changes the README and adds simple CI).

PR #227, opened 2021-07-07
Fix grammar for PackedVectorFormat

kpet:
PackedVectorFormat4x8BitKHR should be enabled by the SPV_KHR_integer_dot_product
extension that first introduced it and not the DotProductInput4x8BitPackedKHR capability,
as per the extension specification.

See http://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/master/extensions/KHR/SPV_KHR_integer_dot_product.html

Signed-off-by: Kevin Petit <kevin.petit@arm.com>

### DIFF
--- a/include/spirv/unified1/spirv.core.grammar.json
+++ b/include/spirv/unified1/spirv.core.grammar.json
@@ -13509,7 +13509,7 @@
         {
           "enumerant" : "PackedVectorFormat4x8BitKHR",
           "value" : 0,
-          "capabilities" : [ "DotProductInput4x8BitPackedKHR" ],
+          "extensions" : [ "SPV_KHR_integer_dot_product" ],
           "version" : "None"
         }
       ]


StuartDBrady (commented):
Looks good in principle - the "enabling capabilities" column for the packed vector format is empty.

Is it better to have done it this way, or does it make no real difference?

What are the semantics for use of PackedVectorFormat4x8BitKHR if the DotProductInput4x8BitPackedKHR capability is not used? _Shouldn't_ such use fail validation?

kpet:
>Is it better to have done it this way, or does it make no real difference?

Well it matches the spec for a start :). I remember we discussed this and concluded that packed vector formats should be easily reusable independently of the dot product capabilities once this goes into core, hence the absence of enabling capability in the spec.

>What are the semantics for use of PackedVectorFormat4x8BitKHR if the DotProductInput4x8BitPackedKHR capability is not used? Shouldn't such use fail validation?

Yes, it definitely should. I found this grammar issue as part of working on extending support in the validator.

StuartDBrady:
> Well it matches the spec for a start :). I remember we discussed this and concluded that packed vector formats should be easily reusable independently of the dot product capabilities once this goes into core, hence the absence of enabling capability in the spec.

In general, I agree.  However, if we added another use of the PackedVectorFormat4x8BitKHR format, this would be under its own capability, which could be made to depend upon the DotProductInput4x8BitPackedKHR capability.  However, that is assuming that if a new extension requires the PackedVectorFormat4x8BitKHR vector format, then the extension would necessarily require the full functionality that DotProductInput4x8BitPackedKHR provides, and perhaps that is not the case.  I'm still a little unclear on this point, unfortunately.

> Yes, it definitely should. I found this grammar issue as part of working on extending support in the validator.

I see... but then I suppose it's not _explicitly_ stated in the specification that this is a validation rule, ("enabled by DotProductInput4x8BitPackedKHR" _implies_ that it's disabled otherwise), but we are saying that _something_ is needed to enable use of the format, and DotProductInput4x8BitPackedKHR is such a capability.

PR #231, opened 2021-07-28
spirv.core.grammar: Remove duplicate OpArbitraryFloatPowNINTEL declaration

MarijnS95:
An identical declaration of `OpArbitraryFloatPowNINTEL` exists just above, with the exact same opcode and operands.

(Context: This breaks one of our code generators which - understandably so - does not perform duplication filtering)


bashbaug (approved):
LGTM

PR #237, opened 2021-09-09
Update CMakeLists.txt

BedrockDotPng:
Fixed CMake file for Shaderc to properly compile.

PR #238, opened 2021-09-10
Fix referenced location for package config file

dneto0:
CMAKE_SOURCE_DIR is the directory containing the top-level CMakeLists.txt file
for the cmake run.  The package file was being referenced as
${CMAKE_SOURCE_DIR}/SPIRV-Headers.pc.in, so if SPIRV-Headers was being
built as part of a larger project, configuration would fail.

Instead, use CMAKE_CURRENT_SOURCE_DIR to reliabley find the pc.in file.


dneto0:
Closing, as this is the same fix as #237

PR #241, opened 2021-09-22
Change the "SpvOpMax" to a 16 bit value (some compilers warn against …

kholtnv:
…comparing 16 bit opcode to the larger 32 bit enumerant).

See 3.37:
Opcode is the low-order 16 bits of word 0 of the instruction, holding its opcode enumerant.


dneto0 (approved):

kholtnv:
Comments from Khronos SPIR-V meeting 9/29.
Perhaps this change may break some implementations because the "Op" enum may now be represented as 16 bits because the MAX enumerant is now 16 bits. (I did build glslangValidator with this change and didn't see any problems.)
Are we limiting the number of enumerants by saying the max is 0x7fff and not 0xffff?

PR #249, opened 2021-10-23
Add `deprecatedVersion` field to annotate phase-out features

PENGUINLIONG:
The current grammar JSON lacks a mechanism to annotate phase-out features that are not currently removed from the specification, which currently include `OpDecorationGroup`, `OpGroupDecorate`, and `OpGroupMemberDecorate` as in SPIR-V 1.5, Revision 5.

This PR introduces a `deprecatedVersion` field to show that such features are not recommended to be emitted and the SPIR-V consumers can decide not to support them, if it's preferred.

`deprecatedVersion` gives the version of SPIR-V specification, in which a feature is deprecated in any of its revisions. It interacts with existing fields `version` and `lastVersion`, where `version` is the first version a feature is supported; and `lastVersion` the version a feature is completely removed from the specification. The lifetime of a feature can thus be bound as following: `version` < `deprecatedVersion` <= `lastVersion`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,7 +29,7 @@
 # https://www.khronos.org/registry/spir-v/
 #
 cmake_minimum_required(VERSION 3.0)
-project(SPIRV-Headers VERSION 1.5.1)
+project(SPIRV-Headers VERSION 1.5.5)
 
 # There are two ways to use this project.
 #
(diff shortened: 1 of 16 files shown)


raunraun:
The SPIR working group likes the idea, but wants to do some internal testing to ensure the various tools that consume the grammar are working with this change.

PENGUINLIONG:
Any update on this?

raunraun:
the group is looking at this right now and will likely have a slight counter proposal that includes multiple fields in the deprecatedVersion tag (such as a link/reference to the replacement functionality the deprecated feature).   Adding fields is usually safe, but there are many consumers of the json that all need testing - which is happening now.

PR #256, opened 2021-11-23
Register TornadoVM SPIRV Beehive Tookit Generator

jjfumero:


jjfumero:
I just saw there is a pending PR requiring the id 32. Once that's merged, I can update to use the ID 33 instead.

jjfumero:
Thank you @raunraun , Let me know if I would need to update the ID to 33 (this PR was before claiming the ID 32: https://github.com/KhronosGroup/SPIRV-Headers/pull/253 )

PR #257, opened 2021-11-24
Atomic float add should not imply Shader

Hardcode84:


antiagainst:
+1. Thanks @Hardcode84! From the [SPV_EXT_shader_atomic_float_add spec](https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/EXT/SPV_EXT_shader_atomic_float_add.asciidoc), `AtomicFloat32AddEXT`/`AtomicFloat64AddEXT` has no implied capabilities, but the commit to update JSON grammar has (https://github.com/KhronosGroup/SPIRV-Headers/commit/7083cb52e1812201c2d5641306fcbca639c9560f). So I think this is a proper fix to make sure the grammar follows the spec.

(We discovered this when trying to use `OpAtomicFAddEXT` for generating OpenCL kernels in MLIR/SPIR-V.)

PR #261, opened 2022-01-04
Include bool type if not already defined

ProkopRandacek:
The bool type is not defined in C by default.

### DIFF
--- a/include/spirv/unified1/spirv.h
+++ b/include/spirv/unified1/spirv.h
@@ -1843,8 +1843,8 @@ typedef enum SpvOp_ {
 } SpvOp;
 
 #ifdef SPV_ENABLE_UTILITY_CODE
-#if !defined(__cplusplus) && !defined(bool)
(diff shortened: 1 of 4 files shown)


raunraun:
@ProkopRandacek - thank you for the fix.  Can you include the headers that get generated by this change as well?  The bots are failing the checks because the generated headers dont match what is in the repo.

ProkopRandacek:
force-pushed the wrong branch, sorry

raunraun:
@ProkopRandacek - It looks like the spirv.h in your commit doesn't match what the tools will generate now, which is causing the bots to fail.  

If you look here: https://github.com/KhronosGroup/SPIRV-Headers#generating-headers-from-the-json-grammar-for-the-spir-v-core-instruction-set  ...you'll see directions to generate all the headers (its about 12 of them for all the different languages).  Can you amend the PR with the headers generated by the tool?

PR #273, opened 2022-04-07
build: use ARCH_INDEPENDENT if possible

Tachi107:
Since this is a header-only library, the CMake Version file should be marked architecture-independent with the [`ARCH_INDEPENDENT`](https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html#command:write_basic_package_version_file) option. As the option has been added in CMake 3.14 and spriv-headers supports CMake 3.0 the option has to be passed only if supported.


dneto0 (approved):
Looks good, thanks!

Issue #276, opened 2022-04-20
Add repack instruction information in grammar file


sjfricke:
As stated in https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4018#discussion_r853691459 it would be nice to have the information about which instructions are [invocation repack instruction](https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#ray-tracing-repack) in the grammar file to be parsed out. Happy to purpose a PR to simply just change the non-useful `"class"  : "Reserved",` into `"class"  : "Repack",` and a definition of the class (if the WG agrees with the direction)

sjfricke:
closed with note from https://github.com/KhronosGroup/SPIRV-Headers/pull/287#discussion_r954518866

PR #287, opened 2022-08-18
Add Invocation Repack Instruction class

sjfricke:
closes https://github.com/KhronosGroup/SPIRV-Headers/issues/276

### DIFF
--- a/include/spirv/unified1/spirv.core.grammar.json
+++ b/include/spirv/unified1/spirv.core.grammar.json
@@ -129,6 +129,10 @@
       "heading" : "Non-Uniform Instructions"
     },
     {
+      "tag"     : "Invocation-Repack",
+      "heading" : "Invocation Repack Instructions"
+    },
+    {
       "tag"     : "Reserved",
       "heading" : "Reserved Instructions"
     }
@@ -4214,7 +4218,7 @@
     },
     {
       "opname" : "OpTraceRayKHR",
-      "class"  : "Reserved",
+      "class"  : "Invocation-Repack",
       "opcode" : 4445,
       "operands" : [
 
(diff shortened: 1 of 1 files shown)


include/spirv/unified1/spirv.core.grammar.json
```diff
@@ -128,6 +128,10 @@
       "tag"     : "Non-Uniform",
       "heading" : "Non-Uniform Instructions"
     },
+    {
+      "tag"     : "Invocation-Repack",
+      "heading" : "Invocation Repack Instructions"
```
jeffbolznv, line 133: We discussed this in the WG call today. This will cause the spec to have a new section named "Invocation Repack Instructions" containing these instructions. We don't think this is appropriate because the invocation repack is more like a side effect of these instructions. If these were to go in a new section, it would probably be a "Ray Tracing Instructions" section. But I think the current state of things is that instructions for extensions stay in Reserved.

Another option might be to add a new field like `"repack" : True`. Or, maybe this logic doesn't need to be in the grammar file at all.

sjfricke: @jeffbolznv thanks for the response

>  This will cause the spec to have a new section 

ouch, was not aware this was being use to generate part of the spec in that way, I agree not worth that

> Another option might be to add a new field like "repack" : True. Or, maybe this logic doesn't need to be in the grammar file at all.

For the scope of this change, happy to just close this as not worth the churn, but... `<rant>`working more with `spirv-val`, the SPIR-V part of the Validation Layers, and other SPIR-V related tools, it is kinda annoying how little information there is to generate from the SPIR-V grammar file in general (compared to the `vk.xml`) and this is an example where for now we are probably fine, but one day there may be a new "repack" instruction added to SPIR-V and people are at the mercy that someone who knows the validation code well knows updates it when an extension comes out (but I also am aware how sometimes extensions don't have the luxury of someone championing who knows all of this).

I feel any effort to add more metadata to the grammar would need to come from internally and assuming no one will have the luxury of time to tackle the problem, maybe worth marking it as a "SPIR-V 2.0" level goal `</rant>`

Issue #303, opened 2022-11-04
for C++11, define & overloads for mask enum types


dneto0:
The overloads for | are already defined, but it would be useful to have & overloads as well.

dneto0:
For example, this exists:

```
inline ImageOperandsMask operator|(ImageOperandsMask a, ImageOperandsMask b) { return ImageOperandsMask(unsigned(a) | unsigned(b)); }  
```

But it would be nice to have the corresponding overload for &

dneto0:
And xor too.

alan-baker:
2022-12-14 WG agreed this is a good idea and should do this.

PR #311, opened 2022-12-14
C++ headers: Define & and ~ bitwise operators for mask enums

dneto0:
Fixes: #303

### DIFF
--- a/tools/buildHeaders/header.cpp
+++ b/tools/buildHeaders/header.cpp
@@ -583,8 +583,7 @@ namespace {
         void printEpilogue(std::ostream& out) const override {
             const Json::Value& enums = spvRoot["spv"]["enum"];
 
-            // Create overloaded operator| for mask types
-            out << "// Overload operator| for mask bit combining\n\n";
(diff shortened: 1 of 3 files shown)


dneto0:
This repo doesn't have unit test support.
I made a tiny smoke test in my copy of SPIRV-Tools to spot check each of the three bitwise operators.

tools/buildHeaders/header.cpp
```diff
@@ -593,8 +592,18 @@ namespace {
                 if (isMask) {
                     const auto typeName = opName + styleStr(enumMask);
 
-                    out << "inline " + typeName + " operator|(" + typeName + " a, " + typeName + " b) { return " +
-                        typeName + "(unsigned(a) | unsigned(b)); }\n";
+                    // Overload operator|
```
alan-baker, line 595: Nit:
```suggestion
                    // Overload bitwise operators.
```

alan-baker (approved):

tools/buildHeaders/header.cpp
dneto0: Each different operator has their own comment.

PR #315, opened 2023-01-17
decouple SPV_KHR_shader_clock from the Shader capability

bashbaug:
Fixes the SPIR-V grammar file to decouple the SPV_KHR_shader_clock extension from the Shader capability.  This coupling was likely from an earlier unpublished version of this extension, and no coupling exists in the published extension specification.

[Here](http://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/KHR/SPV_KHR_shader_clock.html) is the SPV_KHR_shader_clock extension specification, for reference.

Fixes internal issue #721.

### DIFF
--- a/include/spirv/unified1/spirv.core.grammar.json
+++ b/include/spirv/unified1/spirv.core.grammar.json
@@ -4751,7 +4751,6 @@
         { "kind" : "IdScope", "name" : "'Scope'" }
       ],
       "capabilities" : [ "ShaderClockKHR" ],
-      "extensions" : [ "SPV_KHR_shader_clock" ],
       "version" : "None"
     },
     {
@@ -14056,7 +14055,6 @@
         {
           "enumerant" : "ShaderClockKHR",
           "value" : 5055,
-          "capabilities" : [ "Shader" ],
           "extensions" : [ "SPV_KHR_shader_clock" ],
           "version" : "None"
         },


dgkoch (approved):
LGTM

Tobski:
LGTM

Keenuts:
Hello @bashbaug!

I'm currently working on the capability trimming pass, and reached the OpReadClockKHR instruction.

From the linked spec, it seems that to declare the capability, the extension needs to be declared, but this commit removes this. Is that expected?
Uncoupling it from the Shader capability seems valid, is also the extension removal wanted?

Tobski:
Hi @Keenuts, this PR only removed the explicit extension tag on the instruction, not on the capability - the capability enum still requires the extension. Are you seeing an issue in SPIR-V tools resulting from this?

Keenuts:
Hi Tobski!

Yes, that's correct, extension still linked to the capability.

The issue is not yet in SPIRV-Tools, since I haven't finished this pass yet.
What I do is:
 - iterate over all instructions
 - lookup the grammar for this instruction, and gather capabilities + extensions.

I don't do another lookup to look at the extensions listed by the capability itself.
(Maybe I should do that).

Do you know why some instruction like `OpReadClockKHR` doesn't list the extension it needs (since the capability does it), and others like `OpGroupMinNonUniformAMD` do list the `SPV_AMD_shader_ballot` extension?

Tobski:
Yea looking up the capabilities to figure out transitive requirements is intended - there's a lot of inconsistency here because we didn't have a convention or look for it before, and we're currently working through cleaning some of this up. Apparently this is interpreted by some tools as "capability OR extension" as well, so we've decided to do the cleanup along the lines of removing redundant information.

Keenuts:
Right, I see. Thanks 😊
Cleaning redundant information make sense, so I'll use the capability->extension mapping instead.

PR #318, opened 2023-01-28
CMake 3.14

juan-lunarg:
Match Vulkan-Headers 1.3.240 release which `3.14` as the cmake minimum:
- https://github.com/KhronosGroup/Vulkan-Headers

### DIFF
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -9,19 +9,23 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: Install Ubuntu packages
         if: matrix.os == 'ubuntu-latest'
         run: sudo apt install -y dos2unix
       - name: Install macOS packages
         if: matrix.os == 'macos-latest'
         run: brew install dos2unix
+      - name: Setup MSVC environment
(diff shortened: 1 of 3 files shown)


.github/workflows/presubmit.yml
```diff
@@ -10,6 +10,9 @@ jobs:
         os: [ubuntu-latest, macos-latest, windows-latest]
     steps:
       - uses: actions/checkout@v2
+      - uses: lukka/get-cmake@latest
+        with:
+          cmakeVersion: 3.14.0
```
juan-lunarg, line 15: Explicitly test the advertised minimum:
https://discourse.cmake.org/t/check-your-cmake-minimum-required-with-get-cmake/7257

CMakeLists.txt
```diff
@@ -28,7 +28,7 @@
 # The SPIR-V headers from the SPIR-V Registry
 # https://www.khronos.org/registry/spir-v/
 #
-cmake_minimum_required(VERSION 3.0)
+cmake_minimum_required(VERSION 3.14)
```
juan-lunarg, line 31: Vulkan-Headers is also a widely used KhronosGroup library so I'd say it's fine raise the min to 3.14:
https://github.com/KhronosGroup/Vulkan-Headers/blob/9f93cbe76abe9f6cb4a36df10b08fa3b78ae0027/CMakeLists.txt#L17

PR #319, opened 2023-02-01
Update CMake version and project version

juan-lunarg:


CMakeLists.txt
```diff
@@ -28,8 +28,8 @@
 # The SPIR-V headers from the SPIR-V Registry
 # https://www.khronos.org/registry/spir-v/
 #
-cmake_minimum_required(VERSION 3.0)
-project(SPIRV-Headers VERSION 1.5.5)
+cmake_minimum_required(VERSION 3.14)
```
juan-lunarg, line 31: [KhronosGroup/Vulkan-Headers](https://github.com/KhronosGroup/Vulkan-Headers) uses 3.14 so it should be safe for `spirv-headers`. Also `spirv-tools` uses 3.17 as it's minimum

CMakeLists.txt
```diff
@@ -28,8 +28,8 @@
...
+project(SPIRV-Headers VERSION 1.6.1)
```
juan-lunarg, line 32: I was told by @jeremy-lunarg and @greg-lunarg that the project should come from `spirv.core.grammar.json`

juan-lunarg:
@johnkslang / @dneto0

johnkslang:
I'm worried about this because:

- If something is carrying a version number, it should be the version of that thing, not the version of something else, and already (before this PR), the JSON grammar file already wasn't quite right to say its own version number is some SPIR-V version number that gets set independently of its actual rate of updates, which are based on public PRs.  It might be better to not aggrandize this further.

- It seems a bit strange to be using CMAKE to parse a JSON file.  It just automatically tracks some other version number, regardless? Version numbers seem like things that should be bumped up manually after some process verifies the next level of functionality is in place and ready to use, or so that consumers can compare version numbers of the specific component in question (the CMAKE build system?).

Probably the important question is who/what then consumes this parsed version number, and why are they/it not getting it from somewhere else?  (And I probably just didn't look enough to figure that out myself, but it is worth saying in the PR what the change is about, what's driving it, what's accomplished..)

juan-lunarg:
> If something is carrying a version number, it should be the version of that thing, not the version of something else, and already (before this PR)

Originally when I asked @jeremy-lunarg and @greg-lunarg where the version number came from. They told me it came from the `spirv.core.grammar.json`.

That's how I came up with this solution of error checking the project version based off of the json.

jeremy-lunarg:
>     the JSON grammar file already wasn't quite right to say its own version number is some SPIR-V version number that gets set independently of its actual rate of updates

This is my bad. I didn't realize this.

>     * It seems a bit strange to be using CMAKE to parse a JSON file.

The goal was to minimize human error in maintaining the project version. Regardless, it doesn't matter given the previous observation.

.github/workflows/presubmit.yml
```diff
@@ -18,10 +18,8 @@ jobs:
         run: brew install dos2unix
       - name: Build
         run: |
-          mkdir build
-          cd build
-          cmake -DCMAKE_INSTALL_PREFIX=install ..
-          cmake --build . --target install
+          cmake -S . -B build -DCMAKE_INSTALL_PREFIX=install -D SPIRV_HEADERS_SKIP_EXAMPLES=OFF
```
jeremy-lunarg, line 21: FWIW, I liked this line and the update to the cmake minimum version.

PR #321, opened 2023-03-06
Add support for FPGA latency control extension

asudarsa:
This changes add support for the SPIR-V extension documented here: https://github.com/KhronosGroup/SPIRV-Registry/pull/179.
This extension adds two decorations to represent latency controls on the pointer accessed by load, store, pipe read and pipe write instructions.

Thanks


include/spirv/unified1/spirv.core.grammar.json
```diff
@@ -14909,6 +14929,13 @@
           "extensions" : [ "SPV_INTEL_kernel_attributes" ],
           "version" : "None"
         },
+        {
+          "enumerant" : "FPGALatencyControlINTEL",
+          "value" : 6171,
+          "extensions" : [ "SPV_INTEL_fpga_latency_control" ],
+          "extensions" : [ "SPV_INTEL_kernel_attributes" ],
```
bashbaug, line 14936: I don't see the interaction with `SPV_INTEL_kernel_attributes` in the spec.  Is this intended?

If it is intended I think you want a slightly different syntax here:

```suggestion
          "extensions" : [ "SPV_INTEL_fpga_latency_control", "SPV_INTEL_kernel_attributes" ],
```

bashbaug (commented):
I've verified that the enum numbers match what is in the spec.

Mostly LGTM but I found one issue that should be fixed.

include/spirv/unified1/spirv.core.grammar.json
asudarsa: Ah...Good catch. This was not intended. Resolved it. Thanks

Issue #325, opened 2023-03-15
Update to `actions/checkout@v3`


juan-lunarg:
If you look at the actions for CI it's warning about deprecated functionality:

https://github.com/KhronosGroup/SPIRV-Headers/actions/runs/4366849935

> Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: actions/checkout@v2. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.

The fix is very simple. Just update to `actions/checkout@v3`

kpet:
I recommend setting up Dependabot to manage this: https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot. It can be configured to automatically create PRs to migrate actions to their latest version.

Issue #324, opened 2023-03-15
SPIRV-Headers by default builds it's testing


juan-lunarg:
I noticed this while analyzing our build times and found that SPIRV-Headers is building a library by default. Which isn't desirable for package manager systems.

Notably looking around the web multiple package managers have patches against spirv-headers to resolve this exact issue.

https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/5c1bc922781d2a4d62400b8cd42bccbee60956fe/scripts/known_good.json#L30

EDIT:

Conan also has to workaround this minor issue:
https://github.com/conan-io/conan-center-index/blob/e250c399017851f104727aa0aa25b311cfedbdec/recipes/spirv-headers/all/conanfile.py#L30

Spirv-tools as well:
https://github.com/KhronosGroup/SPIRV-Tools/blob/57fb3c7a3f12cd61a26c7ae92fa1d345830307d1/external/CMakeLists.txt#L44

PR #326, opened 2023-03-15
cmake: Overally cleanup/testing update

### LINKED ISSUE #282: pkg-config file is broken when CMAKE_INSTALL_INCLUDEDIR is absolute
As per title: `SPIRV-Headers.pc.in` has
```
prefix=@CMAKE_INSTALL_PREFIX@
includedir=${prefix}/@CMAKE_INSTALL_INCLUDEDIR@
```
and so can’t handle absolute paths in `CMAKE_INSTALL_INCLUDEDIR`. This leads to broken .pc files on NixOS in particular.

Morally identical to https://github.com/KhronosGroup/SPIRV-Tools/issues/3905. See “[Concatenating paths when building pkg-config files](https://github.com/jtojnar/cmake-snips#concatenating-paths-when-building-pkg-config-files)” for a thorough discussion of the problem and a suggested fix, or [KDE’s extra-cmake-modules](https://invent.kde.org/frameworks/extra-cmake-modules/-/blob/master/modules/ECMGeneratePkgConfigFile.cmake#L166) for a simpler approach.

juan-lunarg:
This brings SPIRV-HEADERS up to date with Vulkan-Headers in terms of modern CMake practices/testing.

Minimimum is raised to 3.14 for ARCH_INDEPENDENT functionality, which helps reduces installation of header only libraries. And makes writing/maintaining the CMake much easier.

Lots of testing was added to ensure the stability/interface for package managers. In particular the pkgconfig support is now tested much more extensively for correctness.

The CMake minimum and add_subdirectory workflow is now also explicitly tested for compatibility.

closes #282, #324, #325


jeremy-lunarg (approved):
Awesome! Looks great to me.

juan-lunarg:
Furthermore I'd like to point out this PR is compatible with SPIRV-Tools:

https://github.com/KhronosGroup/SPIRV-Tools/blob/57fb3c7a3f12cd61a26c7ae92fa1d345830307d1/external/CMakeLists.txt#L46

In fact it would be better for spirv-tools since the defaults are much nicer for projects to consume add_subdirectory/find_package.

Since now SPIRV-Tools only installs header files if it's the top level project, and tests are not built by default either.

CMakeLists.txt
```diff
@@ -23,115 +23,43 @@
...
-      COMMAND cmake -E copy_directory ${CMAKE_CURRENT_SOURCE_DIR}/include/spirv
-          $ENV{DESTDIR}${CMAKE_INSTALL_PREFIX}/include/spirv)
-endif()
-
-if (SPIRV_HEADERS_ENABLE_EXAMPLES)
-  message(STATUS "Building SPIRV-Header examples")
-  add_subdirectory(example)
-endif()
+add_library(SPIRV-Headers INTERFACE)
+add_library(SPIRV-Headers::SPIRV-Headers ALIAS SPIRV-Headers)
```
bashbaug, line 35: I'm not a CMake expert, but would it be better for the alias to be `SPIRV::Headers` rather than `SPIRV-Headers::SPIRV-Headers`?  I could imagine at some point we would have other projects in a `SPIRV` namespace.

bashbaug (commented):
I'm not a CMake expert but this looks good to me!  I like the additional testing.

CMakeLists.txt
juan-lunarg: I chose this name to be consistent with the name already being exported by find_package (See tests/find_package for what I mean).  I agree it's not an ideal name, but it's consistent and will cause less potential confusion.

kpet: It's this property that was empty for me https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/8b4a4eb257f88e69eadbf30974680bf8cb6d9b84/layers/CMakeLists.txt#L121. I haven't attempted to get to the bottom of it. Providing `SPIRV_HEADERS_INCLUDE_DIR` externally after commenting out this line was good enough for what I was doing but long term I'd like to be able to use upstream unmodified if possible. I'll try again once this goes in and I cycle back to working on this. In the meantime if someone wants to drive this forward, here's the project structure I'm using (simplified):

* My parent CMake project
  * add_subdirectory(SPIRV-Headers)
  * add_subdirectory(Vulkan-Headers)
  * add_subdirectory(Vulkan-ValidationLayers)

In general it'd be great if the various projects did not assume that their dependencies are installed and could rely on existing targets seamlessly. Having this as a tested config upstream would be grand!

juan-lunarg: > It's this property that was empty for me https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/8b4a4eb257f88e69eadbf30974680bf8cb6d9b84/layers/CMakeLists.txt#L121

Currently the vulkan validation layer code assumes it's finding spirv-headers with `find_package` and as you pointed out will break when using `add_subdirectory` instead. There is nothing that `spirv-headers` can do about this.

Please file an issue in the vulkan-validationlayers and I can fix this assumption to assist your workflow.

> In general it'd be great if the various projects did not assume that their dependencies are installed and could rely on existing targets seamlessly.

The problem with using `add_subdirectory` for grabbing dependencies is that it doesn't play well with system/language package managers (vcpkg, conan, etc) which require `flat builds`.

IE solving this problem:

image

image

juan-lunarg: It may be worth pointing out why VVL focuses on installed dependencies.

Fedora, Debian, Arch Linux, and other distros have vvl as part of their system package management.
- https://salsa.debian.org/xorg-team/vulkan/vulkan-validationlayers/-/blob/debian-unstable/debian/rules

Conan (Language Package manager) does as well.
- https://github.com/conan-io/conan-center-index/tree/master/recipes/vulkan-validationlayers

So they are currently the priority for testing.

Not too mention the SDK packaging for vulkan sdk also utilizes `find_package` / `installation` method due to the aforementioned `flat build` structure benefits that make delivery of multiple libraries/executables significantly simpler / less bug prone.

Apologies this might be more context than needed.

juan-lunarg: Also it's worth pointing out this isn't a CMake exclusive problem. The methodology of building your dependencies as part of your project versus using a package management system is a problem for any C/C++ build system and each build system has their own approach/solution to the problem. And there are pros/cons to either approach.

For now this PR is just a first step towards cleaning things up.

bashbaug: Thanks for the explanations, this makes sense to me.  I missed that we previously installed into the `SPIRV-Headers` namespace (the variables were confusing!) and agree it's best to keep this the same.

I guess if we really want to we can add `SPIRV::Headers` alias later, but that's not a problem we need to solve now.

Thanks for cleaning this up!

bashbaug (approved):

jeremy-lunarg:
@johnkslang I believe the dust has settled. Can you merge this for us?

CMakeLists.txt
dneto0: This is causing Chrome and related projects a lot of grief.

We very much do not want to be forced to use the installed version of a package. We want to point to our own specified set of sources.

PR #330, opened 2023-03-23
Revert "Merge pull request #326 from juan-lunarg/juaramos/cmake"

dneto0:
This reverts commit a41cf67dfa52d33c049dc1b606e2db5cf72dafeb, reversing changes made to 3834d41284f275e8e11f095d5f0fb40ade18abc5.


dneto0:
The cmake cleanup from #326 broke many things in our flows.

dj2:
For our CMake files, we depend on the logic in SPIRV-Tools to find SPIRV-Headers (https://github.com/KhronosGroup/SPIRV-Tools/blob/main/external/CMakeLists.txt#L29). This expects SPIRV-Headers to set a `SPIRV-Headers_SOURCE_DIR` which is what is then used by SPIRV-Tools to find the header files. We do not use `externals/` as the folder to store our `third_party` libraries.

We also do not depend on installed packages, we _always_ use `add_subdirectory` to setup the build.

juan-lunarg:
I'll make a fix adjusting for this, and add testing to CI ensure it's enforced.

dj2:
Awesome, thank you.

juan-lunarg:
See https://github.com/KhronosGroup/SPIRV-Headers/pull/331 for the fix and added testing.

PR #331, opened 2023-03-23
cmake: Overally cleanup/testing update

juan-lunarg:
Added additional testing to cover SPIRV-Tools explicitly


CMakeLists.txt
```diff
@@ -23,115 +23,43 @@
 # CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
 # TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
 # MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
+cmake_minimum_required(VERSION 3.14)
+project(SPIRV-Headers LANGUAGES CXX VERSION 1.5.5)
```
juan-lunarg, line 27: Currently SPIRV-Tools relies on the project name being exactly `SPIRV-Headers`

Testing has been added to make sure this isn't broken.

juan-lunarg (commented):
Tested SPIRV-Tools build process, and VVL build process that consumes both SPIRV-Headers and SPIRV-Tools

jrprice:
Tested this against the project that was broken by the previous PR and can confirm that this works fine now. Thanks for the quick turnaround!

dneto0 (approved):
Thanks for the update.
I verified the delta from #326 was basically only in the place needed to make SPIRV-Tools work again.

PR #333, opened 2023-03-29
docs: Update CMake usage documentation

juan-lunarg:


README.md
```diff
@@ -50,38 +50,31 @@ This includes:
...
-### Using CMake
-A CMake-based project can use the headers without installing, as follows:
+```cmake
+# Example usage of find_package
+add_executable(foobar)
+find_package(SPIRV-Headers REQUIRED CONFIG)
 
-1. Add an `add_subdirectory` directive to include this source tree.
-2. Use `${SPIRV-Headers_SOURCE_DIR}/include}` in a `target_include_directories`
-   directive.
```
juan-lunarg, line 74: NOTE: This is bad practice we shouldn't encourage projects to do. Since it relies on implementation details of `SPIRV-Headers` which shouldn't be relied on.

It's much better to link against the header only library instead.

README.md
```diff
@@ -50,38 +50,31 @@ This includes:
...
-```
-#include "spirv/unified1/GLSL.std.450.h"
-#include "spirv/unified1/OpenCL.std.h"
-#include "spirv/unified1/spirv.hpp"
+target_link_libraries(foobar PRIVATE SPIRV-Headers::SPIRV-Headers)
 ```
 
-See also the [example](example/) subdirectory.  But since that example is
-*inside* this repostory, it doesn't use and `add_subdirectory` directive.
+NOTE: `find_package` works best in tandem with [system package managers](https://en.wikipedia.org/wiki/Package_manager) or language package managers like [conan](https://conan.io/) & [vcpkg](https://learn.microsoft.com/en-us/vcpkg/).
```
juan-lunarg, line 74: NOTE: A reason I mention conan / vcpkg is because SPIRV-Headers is listed as a package of both of these repositories.

- https://conan.io/center/spirv-headers
- https://vcpkg.io/en/packages.html (Search for spirv)

PR #335, opened 2023-04-04
Header support for SPV_EXT_shader_tile_image

janharaldfredriksen-arm:
Header file changes corresponding to https://github.com/KhronosGroup/SPIRV-Registry/pull/198


include/spirv/unified1/spirv.core.grammar.json
```diff
@@ -4102,6 +4102,46 @@
...
+      "opname" : "OpColorAttachmentReadEXT",
+      "class"  : "Image",
+      "opcode" : 4160,
+      "operands" : [
+        { "kind" : "IdResultType" },
+        { "kind" : "IdResult" },
+        { "kind" : "IdRef", "name" : "'Attachment'" },
+        { "kind" : "IdRef", "name" : "'Sample'", "quantifier" : "?" }
+      ],
+      "extensions" : [ "SPV_EXT_shader_tile_image" ],
```
alan-baker, line 4115: For all the instructions remove the extension enable. The desired chain is extension enables capability which enables instruction.

include/spirv/unified1/spirv.core.grammar.json
```diff
@@ -4102,6 +4102,46 @@
...
+      "capabilities" : [ "TileImageColorReadAccessEXT" ],
+      "version" : "1.5"
```
alan-baker, line 4117: This version is only for core spir-v versions. Please remove from all instructions.

include/spirv/unified1/spirv.core.grammar.json
```diff
@@ -10887,6 +10948,13 @@
           "capabilities" : [ "Shader" ],
           "version" : "1.3"
         },
+        {
+          "enumerant" : "TileImageEXT",
+          "value" : 4172,
+          "extensions" : [ "SPV_EXT_shader_tile_image" ],
```
alan-baker, line 10954: Same here.

alan-baker (changes requested):

include/spirv/unified1/spirv.core.grammar.json
janharaldfredriksen-arm: Ack.
Just for my info - when existing ops (OpRayQueryInitializeKHR as an example) _do_ reference the extension name, that's just historical 'mistakes'?

include/spirv/unified1/spirv.core.grammar.json
```diff
@@ -14148,6 +14221,27 @@
           "extensions" : [ "SPV_KHR_8bit_storage" ],
           "version" : "1.5"
         },
+        {
+          "enumerant" : "TileImageColorReadAccessEXT",
+          "value" : 4166,
+          "capabilities" : [ "Shader" ],
```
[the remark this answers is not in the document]
janharaldfredriksen-arm: good catch. We had this in originally but removed it based on feedback (from you, I think). Will remove.

include/spirv/unified1/spirv.core.grammar.json
alan-baker: Yes, we have an issue to try and add this to CI, but haven't gotten around to it yet. This way is better for tooling because it reads both those fields as enables for the instruction. So the parser would accept just having the extension and there would need to be extra code for the capability requirement.

include/spirv/unified1/spirv.core.grammar.json
alan-baker: Sorry I think I messed this up. I believe this should be:
```
"version" : None
```

I believe no entry is interpreted as 1.0.

include/spirv/unified1/spirv.core.grammar.json
janharaldfredriksen-arm: OK. Thanks for the info.

include/spirv/unified1/spirv.core.grammar.json
janharaldfredriksen-arm: Aha. No worries. Added version fields back - with 'None'.

include/spirv/unified1/spirv.core.grammar.json
```diff
@@ -11061,6 +11122,12 @@
           "enumerant" : "SubpassData",
           "value" : 6,
           "capabilities" : [ "InputAttachment" ]
+        },
+        {
+          "enumerant" : "TileImageDataEXT",
+          "value" : 4173,
+          "capabilities" : [ "TileImageColorReadAccessEXT" ],
+          "version" : "None"
```
janharaldfredriksen-arm, line 11130: I was not sure if I should add a version field here or not (since these enumerants haven't been extended before). I saw the StorageClass list above did include it so did the same, though the ImageFormat further down does not. Let me know if there's a preference for having it here vs not.

dneto0 (approved):
Thanks!

include/spirv/unified1/spirv.core.grammar.json
janharaldfredriksen-arm: This was updated in the most recent commit.

alan-baker (approved):

PR #336, opened 2023-04-06
Shift Codeplay opcode range to preserve alignment of opcodes and non-opcodes

DBDuncan:
We wish to preserve our opcode and non-opcode alignment but without reserving a space we probably won't use. Don't want to put nothing otherwise it would be very difficult to spot. But another style might be better. Perhaps a comment instead.


johnkslang:
The reason we have separated these out is so that they can be allocated independently.  All enums can be in principle reuse the same numbers, but, yes, for the simplicity desired here, we are currently using a convention of not requiring reusing the numbers across different enums.

The exception to that is for opcodes.  That is because while we have ~4 billion values for the other enumerants, we only have ~65,000 for opcodes, and hence don't want to tie these together.  It is better to more slowly consume the opcodes.  This is the reason spir-v.xml distinguishes between opcodes and other enumerants.

So, I'm pretty unclear that this change actually needs to be made.

Hugobros3:
My understanding is that the potential for misalignment was deliberate, as otherwise there is no need for two tables

PR #337, opened 2023-04-13
Add NonSemantic.AuxData

sarnex:
This non-semantic EIS is used to specify preserving otherwise unhandled function metadata and attributes.


Hugobros3:
What kind of metadata and attributes is this intended to be for ?

We have some specific function debug info already in [NonSemantic.DebugInfo.100](https://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/main/nonsemantic/NonSemantic.Shader.DebugInfo.100.html#DebugFunction) and there's a [new version](https://github.com/KhronosGroup/SPIRV-Registry/pull/186) being worked on as well, how does this relate ?

PR #338, opened 2023-04-13
Add Hero C Compiler to the vendor list & add C source language

heroseh:


alan-baker:
It looks like you haven't regenerated the headers (e.g. spirv.h). See the instructions [here](https://github.com/KhronosGroup/SPIRV-Headers#generating-headers-from-the-json-grammar-for-the-spir-v-core-instruction-set).

include/spirv/unified1/spirv.core.grammar.json
```diff
@@ -10065,6 +10065,10 @@
         {
           "enumerant" : "SYCL",
           "value" : 7
+        },
+        {
+          "enumerant" : "C",
```
Hugobros3, line 10070: This addition was discussed during the last SPIR-V call today, here's a summary:

We have already various language entries for OpenCL C and C++ flavours, so perhaps "C" is a bit too generic. Both in the sense that it appears you are implementing a subset of C, but also you have your own extensions for exposing things like textures etc which do not exist in standard C. Perhaps it makes sense to define your own language for your dialect of C, and create a language entry for that ?

bashbaug:
Ping @heroseh - just want to be sure you saw the comments above.  Let us know if there is anything we can do to help.  Thanks!

heroseh:
Sorry for the delayed response, thank you all for taking the time to review this! I have made the requested adjustments. Is it all good now?

PR #346, opened 2023-05-05
Make the generated operators for masks constexpr

kpet:
Generate the overloaded C++11 operators as constexpr instead of inline. constexpr implies inline but makes it possible to use the operators in constexpr functions.


johnkslang:
> Generate a MaskAll mask value that is the union of all the values defined. This provides a ready-to-use definition for "all valid mask values" that a user can test against.

Wouldn't the user want to know this relative to which extensions are enabled/supported, and possibly not have the value change with a header update just because, say, a vendor reserved another token?

kpet:
Discussed in the 2023/05/17 teleconference: concerns about introducing a `MaskAll` definition for the various enums. This used to exist and created issues. I've now removed it from this PR.

PR #351, opened 2023-06-15
Support for VK_KHR_shader_quad_scope

ruiminzhao:


ruiminzhao:
@johnkslang Hello John Kessenich, Is there anyway to delete this PR as it has some private info, thanks!

PR #355, opened 2023-06-21
Add definitions for SPV_KHR_cooperative_matrix

kpet:


include/spirv/unified1/spirv.core.grammar.json
```diff
@@ -4493,6 +4493,77 @@
...
       "version" : "1.6"
     },
+    {
+      "opname" : "OpTypeCooperativeMatrixKHR",
+      "class"  : "Type-Declaration",
+      "opcode" : 4456,
+      "operands" : [
+        { "kind" : "IdResult" },
+        { "kind" : "IdRef",        "name" : "'Component Type'" },
+        { "kind" : "IdScope",      "name" : "'Execution'" },
```
alan-baker, line 4503: Nit: https://github.com/KhronosGroup/SPIRV-Registry/pull/207 calls this "Scope".

include/spirv/unified1/spirv.core.grammar.json
```diff
@@ -4493,6 +4493,77 @@
...
+    {
+      "opname" : "OpCooperativeMatrixStoreKHR",
+      "class"  : "Memory",
+      "opcode" : 4458,
+      "operands" : [
+        { "kind" : "IdRef",             "name" : "'Pointer'" },
+        { "kind" : "IdRef",             "name" : "'Object'" },
+        { "kind" : "IdRef",             "name" : "'MemoryLayout'" },
+        { "kind" : "IdRef",             "name" : "'Stride'", "quantifier": "?" },
+        { "kind" : "MemoryAccess",      "quantifier" : "?" }
```
alan-baker, line 4535: Nit: https://github.com/KhronosGroup/SPIRV-Registry/pull/207 calls this "Memory Operand"

include/spirv/unified1/spirv.core.grammar.json
```diff
@@ -4493,6 +4493,77 @@
...
+      "opname" : "OpCooperativeMatrixMulAddKHR",
+      "class"  : "Arithmetic",
+      "opcode" : 4459,
+      "operands" : [
+        { "kind" : "IdResultType" },
+        { "kind" : "IdResult" },
+        { "kind" : "IdRef",             "name" : "'A'" },
+        { "kind" : "IdRef",             "name" : "'B'" },
+        { "kind" : "IdRef",             "name" : "'C'" },
+        { "kind" : "CooperativeMatrixOperands",             "quantifier" : "?" }
```
alan-baker, line 4550: Nit: This should be named "Cooperative Matrix Operands"

alan-baker (approved):
Just some small naming inconsistencies between the spec and the headers.

include/spirv/unified1/spirv.core.grammar.json
kpet: Done

include/spirv/unified1/spirv.core.grammar.json
kpet: Done

include/spirv/unified1/spirv.core.grammar.json
kpet: I've added the name so the headers are now aligned to the spec but I think the spec and grammar for this operand are actually wrong. They don't allow for additional operands after the initial Memory Operand literal but additional operands are required for **MakePointerAvailable**/**MakePointerVisible** which were definitely intended to be usable. This means none of the CTS or SPIR-V Tools tests (or glslang) covered their use.

kpet: Correction: SPIR-V Tools validation tests seem to be exercising this and we are using a "?" quantifier for OpLoad/OpStore so the grammar is in line with other instructions taking Memory Operands and it seems that for most operand kinds we translate both the "?" and "\*" quantifiers to a single "optional" flag. This is really confusing, one would expect "?" to mean "zero or one" and "*" to mean "zero or more". The spec language does need changing though.

kpet: Finally reached the end of the rabbit hole (and definitely past the point where I should have stopped for the day). The grammar tables translate "?" and "*" as expected and Memory Access operands are special-cased to handle additional operands. The grammar and tooling are all fine.

alan-baker: The way MakePointerAvailable/Visible are handled in the tooling is that those bits cause another operand to be required.

kpet: Yup, it's all fine and all makes sense. I was tired enough to enter a confusion spiral :).

johnkslang (approved):

PR #358, opened 2023-07-06
Add WGSL source language

alan-baker:
See https://www.w3.org/TR/WGSL/


dneto0 (approved):
I believe the canonical URL for the spec is https://www.w3.org/TR/WGSL/   :-)

dneto0 (approved):

bashbaug (approved):

PR #359, opened 2023-07-11
cmake: Use share instead of DATADIR for CMake config files

juan-lunarg:
See the following discussion for context:
https://discourse.cmake.org/t/what-should-the-destination-be-for-a-header-only-librarys-cmake-config-file/8473/2

NOTE: I've never seen this cause an issue before, but better safe than sorry. Also I think it reads a bit better.

I recently did this for VulkanHeaders so I'd like this fixed here as well.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,15 +45,12 @@ if (PROJECT_IS_TOP_LEVEL)
     include(CMakePackageConfigHelpers)
 
     install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/include/spirv DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
-
-    set(cmake_install_dir "${CMAKE_INSTALL_DATADIR}/cmake/SPIRV-Headers")
-    set(version_config "${CMAKE_CURRENT_BINARY_DIR}/generated/SPIRV-HeadersConfigVersion.cmake")
-
-    write_basic_package_version_file("${version_config}" COMPATIBILITY SameMajorVersion ARCH_INDEPENDENT)
-    install(FILES "${version_config}" DESTINATION "${cmake_install_dir}")
-
     install(TARGETS SPIRV-Headers EXPORT "SPIRV-HeadersConfig" INCLUDES DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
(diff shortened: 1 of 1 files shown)


dneto0:
I see what you're trying to do.

But this seems contrary to the rule / pattern for GNUInstallDirs.

See https://cmake.org/cmake/help/v3.0/module/GNUInstallDirs.html:
     * DATADIR is for "read-only architecture-independent data (DATAROOTDIR)"

Looking up 'datadir' in https://www.gnu.org/prep/standards/html_node/Directory-Variables.html, it says

> This should normally be /usr/local/share, but write it as $(datarootdir). ...
> The definition of ‘datadir’ is the same for all packages, so you should install your data in a subdirectory thereof. Most packages install their data under $(datadir)/package-name/.

Indeed the default sets the path component to "share". See https://github.com/Kitware/CMake/blob/ff3179cef8bbf004dade5140a23bdd23e1c06073/Modules/GNUInstallDirs.cmake#L316 

But it's also intended to be overridden.  https://github.com/Kitware/CMake/blob/ff3179cef8bbf004dade5140a23bdd23e1c06073/Modules/GNUInstallDirs.cmake#L163 

So I think the answer in the discourse is a particular instantiation, where the person configuring the project has not overridden CMAKE_INSTALL_DATADIR.  But the whole point of supporting GNUInstallDirs is to have good defaults *and* to allow overriding.

So I think this PR is actually a regression.

dneto0 (changes requested):
I think this is a regression, because it takes away the ability to override with CMAKE_INSTALL_DATADIR.

More complete comments at https://github.com/KhronosGroup/SPIRV-Headers/pull/359#issuecomment-1633146269

juan-lunarg:
If you see the discussion one of the CMake maintainers [comments](https://discourse.cmake.org/t/what-should-the-destination-be-for-a-header-only-librarys-cmake-config-file/8473/2?u=buildsystemperson):

> I was originally tempted to use ${CMAKE_INSTALL_DATADIR} in my previous answer too, but that’s not what you want. The path searched by find_package() is share, it can’t be something else.

IE if someone set `CMAKE_INSTALL_DATADIR` to `shared` that would actually break find_package support.

See the find_package documentation for reference.

PR #360, opened 2023-07-19
Report failures in makeHeaders

kpet:
Exit the script when one of the called command fails.

### DIFF
--- a/tools/buildHeaders/bin/makeHeaders
+++ b/tools/buildHeaders/bin/makeHeaders
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+set -ex
+
 python3 bin/makeExtinstHeaders.py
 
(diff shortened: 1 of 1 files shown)


bashbaug (approved):

johnkslang:
It looks like we have a `LiteralFloat` in the grammar file, from June 21st that should be fixed, regarding `FPMaxErrorDecorationINTEL`.

@bashbaug

bashbaug:
This should be good now that we have reverted the `LiteralFloat` changes, but if we want to be sure we can rebase this PR and these tests should pass.

kpet:
@bashbaug I've rebased the change. I agree it would be good to see it pass now that we've validated that it does enable the CI to fail when there are generation issues.

bashbaug:
CI is clean and there are two approvals.  Merging.

PR #361, opened 2023-07-19
Remove capabilities on cooperative matrix enums to align with specifcation

kpet:


dneto0:
This change broke basic parsing of these operands in SPIRV-Tools.

Taking the capability away means the operand is no longer visible to the parsers:

      // We consider the current operand as available as long as
      // 1. The target environment satisfies the minimal requirement of the
      //    operand; or
      // 2. There is at least one extension enabling this operand; or
      // 3. There is at least one capability enabling this operand.
    
https://github.com/KhronosGroup/SPIRV-Tools/blob/1d14d84f291805ce845a0e5b9775e5e0ab11995b/source/operand.cpp#L66

I think the fix is then to *add* a dependence on the extension.  I'll make that change.

dneto0:
(Also, unfortunately, the SPIRV-Tools unit tests did not exercise the "Cooperative Matrix Operands" operand on the mul-add instruction, so this wasn't caught until trying to validate a test case in Glslang.  Glslang generates correct code but the spir-v validator can't parse it because of this grammar snafu)

PR #367, opened 2023-08-04
Cooperative matrix enums depend on the extension

dneto0:
This fixes basic parsing in SPIRV-Tools.

An enum is considered invisible by SPIRV-Tools if:
  - it is not in any SPIR-V core version
  - it has no enabling capabilities
  - it has no enabling extensions

So make the new cooperative matrix enums depend on the extension, since they are not in any core version of SPIR-V, and don't depend on any capabilities.

### DIFF
--- a/include/spirv/unified1/spirv.core.grammar.json
+++ b/include/spirv/unified1/spirv.core.grammar.json
@@ -15390,31 +15390,37 @@
         {
           "enumerant" : "NoneKHR",
           "value" : "0x0000",
+          "extensions" : [ "SPV_KHR_cooperative_matrix" ],
           "version" : "None"
         },
         {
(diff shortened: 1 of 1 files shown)


dnovillo (approved):
Nice catch!

alan-baker:
This was intentional. #361 removed the dependence on the capability. These enums can only appear in instructions guarded by the capability so there is a transitive dependence.

dneto0:
> This was intentional. #361 removed the dependence on the capability. These enums can only appear in instructions guarded by the capability so there is a transitive dependence.

Thanks to your pushback I've got a PR in flight to fix SPIRV-Tools' behaviour here.
See https://github.com/KhronosGroup/SPIRV-Tools/pull/5370 

We should be able to revert this change.

Issue #368, opened 2023-08-04
add CI check: each enum is visible


dneto0:
To avoid problems like https://github.com/KhronosGroup/SPIRV-Headers/pull/367, add a check:

Each enum in the grammar should be "visible" in the sense SPIRV-Tools needs it. It must satisfy one of:
- it's in a core SPIR-V revision, or
- it's enabled by a capability, or
- it's enabled by an extension

Issue #278, opened 2022-04-20
CI headers check should ensure non-capability enums are guarded by capability, not by extension


dneto0:
When making a new extension, the typical best practice is that:
- a new capability is introduced, and that is guarded by the extension
- other new tokens: instructions and enums, are guarded by that new enum alone.

The CI check should check this. It's the most common corrective feedback given when reviewing new extensions.

PR #369, opened 2023-08-04
Validate enums have a sensible versions and are visible

dneto0:
Fixes: #368

### DIFF
--- a/tools/buildHeaders/jsonToSpirv.cpp
+++ b/tools/buildHeaders/jsonToSpirv.cpp
@@ -36,7 +36,49 @@
 
 #include "jsonToSpirv.h"
 
+namespace {
+// Returns true if the given string is a valid SPIR-V version.
+bool validSpirvVersionString(const std::string s) {
+  return
+  s == "1.0" ||
+  s == "1.1" ||
+  s == "1.2" ||
+  s == "1.3" ||
+  s == "1.4" ||
+  s == "1.5" ||
+  s == "1.6";
+}
+
+// Returns true if the given string is a valid version
+// specifier in the grammar file.
+bool validSpirvVersionStringSpecifier(const std::string s) {
+  return s.empty() || s == "None" || validSpirvVersionString(s);
+}
+}  // anonymous namespace
+
 namespace spv {
+
+bool EnumValue::IsValid(const std::string& context) const
+{
+  bool result = true;
+  if (!validSpirvVersionStringSpecifier(firstVersion)) {
+    std::cerr << "Error: " << context << " " << name << " \"version\" is invalid: " << firstVersion << std::endl;
+    result = false;
+  }
+  if (!validSpirvVersionStringSpecifier(lastVersion)) {
+    std::cerr << "Error: " << context << " " << name << " \"lastVersion\" is invalid: " << lastVersion << std::endl;
+    result = false;
+  }
+
+  // It must be visible.
+  const bool visible = (firstVersion != "None") || !extensions.empty() || !capabilities.empty();
+  if (!visible) {
+    std::cerr << "Error: " << context << " " << name << " is not visible: "
+              << "its version is set to \"None\", and it is not enabled by a capability or extension" << std::endl;
+    result = false;
+  }
+  return result;
+}
 
 // The set of objects that hold all the instruction/operand
 // parameterization information.
@@ -284,6 +326,8 @@ void jsonToSpirv(const std::string& jsonPath, bool buildingHeaders)
         return;
     initialized = true;
 
+    size_t errorCount = 0;
+
     // Read the JSON grammar file.
     bool fileReadOk = false;
     std::string content;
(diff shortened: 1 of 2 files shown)


dneto0:
I have a PR in flight that relaxes SPIRV-Tools' parsing, so we don't need to be as strict as this.  I'll modify this PR in a corresponding way.

tools/buildHeaders/jsonToSpirv.cpp
```diff
@@ -36,8 +36,58 @@
...
+{
+  bool result = true;
+  if (firstVersion.empty()) {
+    std::cerr << "Error: " << context << " " << name << " \"version\" must be set, probably to \"None\"" << std::endl;
+    result = false;
+  } else if (!validSpirvVersionStringSpecifier(firstVersion)) {
+    std::cerr << "Error: " << context << " " << name << " \"version\" is invalid: " << firstVersion << std::endl;
+    result = false;
+  }
+  if (!lastVersion.empty() && !validSpirvVersionStringSpecifier(lastVersion)) {
```
alan-baker, line 71: I don't think we'd want to see `lastVersion: None` so how about disallowing that case?

```suggestion
  if (!lastVersion.empty() && !validSpirvVersionString(lastVersion)) {
```

tools/buildHeaders/jsonToSpirv.cpp
```diff
@@ -36,8 +36,58 @@
...
+    std::cerr << "Error: " << context << " " << name << " \"lastVersion\" is invalid: " << lastVersion << std::endl;
+    result = false;
+  }
+
+  if (oc == OperandCapability || oc == OperandOpcode) {
```
alan-baker, line 76: Worth a comment about why only capabilities and opcodes are checked here.

include/spirv/unified1/spirv.core.grammar.json
```diff
@@ -14191,7 +14862,8 @@
         {
           "enumerant" : "CoreBuiltinsARM",
           "value" : 4165,
-          "extensions" : [ "SPV_ARM_core_builtins" ]
+          "extensions" : [ "SPV_ARM_core_builtins" ],
+          "version": "None"
```
alan-baker, line 14866: Nice, shows this is worth doing.

alan-baker (changes requested):
A couple small suggestions.

tools/buildHeaders/jsonToSpirv.cpp
dneto0: fixed.

alan-baker (approved):

bashbaug:
WRT #278, a pretty common mistake is to guard an instruction by both a capability and its extension, for example (this is just an example, it's not in the actual json file):

```json
    {
      "opname" : "OpControlBarrierArriveINTEL",
      "class"  : "Barrier",
      "opcode" : 6142,
      "operands" : [
        { "kind" : "IdScope",           "name" : "'Execution'" },
        { "kind" : "IdScope",           "name" : "'Memory'" },
        { "kind" : "IdMemorySemantics", "name" : "'Semantics'" }
      ],
      "capabilities" : [ "SplitBarrierINTEL" ],
      "extensions" : [ "SPV_INTEL_split_barrier" ],
      "version" : "None"
    },
```

I tried modifying the json file to do this and it didn't get flagged as an error.  Actually, it doesn't seem to get flagged by an error if I only include the extension and do not include a capability.  Is this intended behavior?

dneto0:
> WRT https://github.com/KhronosGroup/SPIRV-Headers/issues/278, a pretty common mistake is to guard an instruction by both a capability and its extension, for example (this is just an example, it's not in the actual json file):

There are 78  cases like this already, e.g. OpSubgroupBallotKHR.
I intend to grandparent them in with an allowlist.

dneto0:
I also fixed the logic for checking capabilites. See the commit message (and the comments in the patch)

alan-baker (approved):

tools/buildHeaders/jsonToSpirv.cpp
```diff
@@ -36,8 +36,185 @@
...
+// Returns true if the given string is a valid version
+// specifier in the grammar file.
+bool validSpirvVersionStringSpecifier(const std::string s) {
+  return s == "None" || validSpirvVersionString(s);
+}
+}  // anonymous namespace
+
 namespace spv {
 
+bool IsLegacyDoublyEnabledInstruction(const std::string& instruction) {
```
bashbaug, line 61: I suspect you've considered this, but just to ask the question: it's too late to change these in the json file?

bashbaug (approved):
Thanks!  This is a nice improvement.

I verified that the example I included previously that guards an instruction by both a capability and an extension is flagged an an error now. 👍 

Should we also flag instructions that are guarded by an extension but not a capability as an error, for example (again, this is just an example, it's not actually in the json file):

```json
    {
      "opname" : "OpControlBarrierArriveINTEL",
      "class"  : "Barrier",
      "opcode" : 6142,
      "operands" : [
        { "kind" : "IdScope",           "name" : "'Execution'" },
        { "kind" : "IdScope",           "name" : "'Memory'" },
        { "kind" : "IdMemorySemantics", "name" : "'Semantics'" }
      ],
      "extensions" : [ "SPV_INTEL_split_barrier" ],
      "version" : "None"
    },
```

I'm fine merging this as-is; I mostly want to understand the policy and track the additional checks, assuming they're something we want to do.

PR #370, opened 2023-08-04
Enforce best practices for guarding enums

dneto0:
Enforce best practices for guarding enums
    
    An enum introduced by an extension (that is not in a core SPIR-V
    version) is either:
    - a capability, or
    - not a capability, but guarded by a capability and not guarded by an extension.
    
    Allow cases that have already landed, with an allow-list.
    
Fixes: #278

Builds on #369


dneto0:
Folded into #369

PR #372, opened 2023-08-16
Allow OpMatrixTimesScalar for cooperative matrix type

vmaksimo:
This change fixes `spirv-val` validation error, when we use `OpMatrixTimesScalar` with the `TypeCooperativeMatrixKHR`, allowing to use one of the expected capabilities instead of only `Matrix` one:

```
Opcode MatrixTimesScalar requires one of these capabilities: Matrix
```

Extension spec: https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/KHR/SPV_KHR_cooperative_matrix.asciidoc

### DIFF
--- a/include/spirv/unified1/spirv.core.grammar.json
+++ b/include/spirv/unified1/spirv.core.grammar.json
@@ -1547,7 +1547,10 @@
         { "kind" : "IdRef",        "name" : "'Matrix'" },
         { "kind" : "IdRef",        "name" : "'Scalar'" }
       ],
-      "capabilities" : [ "Matrix" ]
+      "capabilities": [
+        "Matrix",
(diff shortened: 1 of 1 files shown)


MrSidims (approved):
Thanks!

bashbaug (approved):
Nice catch!

bashbaug:
> In environments where it can't, **OpMatrixTimesScalar** cannot be used.

Ah, right, I'd forgotten this was how we resolved this issue, even though I left a "note for future me" 😄 

I think this PR can be closed (not merged).

bashbaug:
Closing as discussed in the August 23rd teleconference (and see discussion above).

PR #380, opened 2023-09-17
Add LiteralFloat to operand_kinds

rAzoR8:
LiteralFloat was introduced with FPMaxErrorDecorationINTEL, but no entry in operand_kinds corresponds to it. Some codegen tooling might depend on operand_kinds -> category mapping.

This PR adds it with the following "doc" string: "A float consuming one word"
Maybe better wording is necessary.

```
{
      "category" : "Literal",
      "kind" : "LiteralFloat",
      "doc" : "A float consuming one word"
}
```

Best,
Fabian

### DIFF
--- a/include/spirv/unified1/spirv.core.grammar.json
+++ b/include/spirv/unified1/spirv.core.grammar.json
@@ -16483,6 +16483,11 @@
     },
     {
       "category" : "Literal",
+      "kind" : "LiteralFloat",
+      "doc" : "A float consuming one word"
+    },
+    {
+      "category" : "Literal",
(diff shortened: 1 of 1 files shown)


bashbaug (approved):
LGTM, this is definitely an improvement.

I suppose it's possible in the future this could be "a float consuimg one or more words", similar to "LiteralInteger", if we start supporting 64-bit float literals?  I think all "LiteralFloats" are currently 32-bits and hence one word, though.

bashbaug:
Merging as discussed in the September 20th teleconference.

PR #386, opened 2023-10-20
Make C++ operator overload for enums not happen for HLSL

### LINKED ISSUE #385: Surround C++ operator overloads in hpp and hpp11 with an `#ifndef SPV_DISABLE_ENUM_OPERATORS `
Hi, we're doing a slightly weird thing, which is we're including the SPIR-V Header in our HLSL code to make the writing on SPIR-V intrinsics easier (don't need to remember the opcodes)

However HLSL doesn't support global operator overloads, so we need a define similar to `SPV_ENABLE_UTILITY_CODE` (although a negative one so backwards compatibility it preserved)

alternative is `#ifndef __HLSL_VERSION` which checks for the HLSL compiler.

devshgraphicsprogramming:
This allows for inclusion of the headers in HLSL2021 shaders.

Helps with maintenance of large headers of SPIR-V intrinsics (don't need to type out the opcodes by hand).

Fixes #385


dneto0 (changes requested):
It would be better to generate a new file specific to HLSL.

devshgraphicsprogramming:
> It would be better to generate a new file specific to HLSL.

Given the direction HLSL is heading, I'll eventually expect the C++11 file to work without issues.

bashbaug:
Hi @devshgraphicsprogramming, just checking in to ensure you saw the review comment above and aren't waiting on anything from us.  Thanks!

devshgraphicsprogramming:
> Hi @devshgraphicsprogramming, just checking in to ensure you saw the review comment above and aren't waiting on anything from us. Thanks!

And I responded, wonder if you saw that.

bashbaug:
> And I responded, wonder if you saw that.

I think so, do you mean your response:

> I'll eventually expect the C++11 file to work without issues.

If this is the case, is this PR needed?  Thanks!

devshgraphicsprogramming:
> > And I responded, wonder if you saw that.
> 
> I think so, do you mean your response:
> 
> > I'll eventually expect the C++11 file to work without issues.
> 
> If this is the case, is this PR needed? Thanks!

The HLSL 202x project has not committed to this feature, nobody knows if and when they'll add operator overloads. Its probably contingent on proposal 0006 and thats probably years away.

A PR is needed either with this or HLSL specific headers which is what I suggested in #385 but got no answer.

bashbaug:
Let's do this by generating a specific file for HLSL, as @dneto suggested [above](https://github.com/KhronosGroup/SPIRV-Headers/pull/386#pullrequestreview-1698282894).  This way the C++ files remain unchanged and if any other changes are needed for HLSL they can be easily added.

@devshgraphicsprogramming can you please update this PR to generate an HLSL-specific file?  Thanks!

PR #395, opened 2023-12-06
Add a few missing calls to std::exit on error

kpet:
Errors were printed but did not stop processing and as a result the related issues were not caught by CI checks.

Change-Id: I030c7ffe23a4730106a19e212895bef03b6ca456


bashbaug (approved):
Merging as discussed in the December 6th teleconference.

PR #397, opened 2023-12-27
Enable Dependabot to keep CI up to date

### LINKED ISSUE #396: Enable Dependabot to keep CI up to date
I'd like to suggest to enable dependabot for keeping the actions used on presubmit.yml workflow in case of new major versions. It is considered a security practice to be up to date to new versions in order to get vulnerability and bug fixes.

Since the presubmit.yml file is a read only workflow, with no access to secrets, I don't think we need to hash pin it --- we can keep it major version pinned. 

Let me know what you think about enabling dependabot for this workflow to get major version patches. I'll submit a PR with the dependabot configuration file together with this issue.

Thanks!

joycebrum:
Closes #396 

Here is a configuration file for dependabot.

I decided to configure it group updates on a single PR in order to avoid multiple PRs. Although, since it is major version updates, single PRs would probably since it major versions are not releases too often.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    groups:
+      github-actions:
+        patterns:
+          - "*"


bashbaug (approved):
Merging as discussed in the January 3rd teleconference.

PR #398, opened 2023-12-28
Publish the header for the vulkan-shader-profiler embedded reflection…

rjodinchr:
… non-semantic instruction set

vulkan-shader-profiler is a tool based on perfetto to help vulkan developer profil their applications.
First on the Host side using perfetto, but then using the extractor and the runner, this tools allow to get inside about a shader using counters (based on SPV_KHR_shader_clock)

### DIFF
--- a/include/spirv/unified1/NonSemanticVkspReflection.h
+++ b/include/spirv/unified1/NonSemanticVkspReflection.h
@@ -1,0 +1,57 @@
+// Copyright (c) 2020 The Khronos Group Inc.
+// 
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and/or associated documentation files (the
+// "Materials"), to deal in the Materials without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Materials, and to
+// permit persons to whom the Materials are furnished to do so, subject to
+// the following conditions:
+// 
+// The above copyright notice and this permission notice shall be included
+// in all copies or substantial portions of the Materials.
+// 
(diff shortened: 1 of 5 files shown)


dneto0 (approved):
LGTM.
SPIR WG will do final approval.

include/spirv/unified1/NonSemanticVkspReflection.h
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) 2020 The Khronos Group Inc.
```
bashbaug, line 1:
```suggestion
// Copyright (c) 2024 The Khronos Group Inc.
```

We need to update other copyright dates in this repo also, but we may as well get this one right from the start.

bashbaug (commented):
Is the spec for this extended instruction set posted anywhere?  I don't see a SPIRV-Registry PR or similar.  It's not strictly required, but I can't really comment on the content of these changes without a spec to compare against - thanks!

include/spirv/unified1/NonSemanticVkspReflection.h
rjodinchr: I did not remember why I did not put 2024 at the time.
Commit your change reminds me of the reason.:
It breaks the formatting check as this file needs to be the exact one generated from https://github.com/KhronosGroup/SPIRV-Headers/blob/main/tools/buildHeaders/bin/generate_language_headers.py#L35 which has `2020` as the date.

rjodinchr:
> Is the spec for this extended instruction set posted anywhere? I don't see a SPIRV-Registry PR or similar. It's not strictly required, but I can't really comment on the content of these changes without a spec to compare against - thanks!

I don't intend to have anything in the SPIRV-Registry for that extended instruction set.

bashbaug (approved):
Merging as discussed in the January 10th teleconference.

PR #399, opened 2024-01-01
Register SPIR-Q tools for SPIR-V

PENGUINLIONG:
Currently we only have an assembler but there will be a code generator soon.


AnnThorsnes:
Hello, my name is Ann Thorsnes from the Khronos Group. We noticed that your GitHub project called SPIR-Q is very similar to our trademarked names SPIR™ and SPIR-V™, and we are concerned about potential confusion. We would really appreciate if you could change the name of your project to avoid any confusion?  Thanks so much for your help in this matter! (ann.thorsnes@khronosgroup.org)

PENGUINLIONG:
@AnnThorsnes Well I saw there were tools with the SPIR- prefix in the list. I thought it was alright. I hope the lowercase version, `spirq`, is possible because many some other projects have been depending on the package for years.

AnnThorsnes:
Hello - thanks for your response and sorry for the delay in getting back to you. We do ask that you don't use our trademarked name. Can you change the name to SPQ?

PENGUINLIONG:
@AnnThorsnes Sorry about that, it's now spq.

AnnThorsnes:
Great - thank you! If you can change here as well: https://github.com/PENGUINLIONG/spirq-rs

PENGUINLIONG:
@AnnThorsnes Made another tweak.

bashbaug (approved):
Merging as discussed in the March 20th teleconference.

PR #405, opened 2024-01-17
Register Zig Compiler tool

Snektron:
Registers the Zig Compiler as SPIR-V tool


bashbaug (approved):
Merging as discussed in the January 17th teleconference.

PR #406, opened 2024-01-24
reserve one scope bit for an Intel extension

bashbaug:
Please reserve one "scope" bit for an Intel extension.

This appears to be the first time a non-Khronos extension has needed a scope bit, so this is a new section in the SPIR-V XML file.  I tried to follow existing conventions for other bit reservations, but please let me know if I missed something.

Please pay careful attention to:

1. I've kept the low 16 bits reserved for Khronos use, similar to the other bit reservations.  Do we still want to do this?
2. I did NOT keep the highest bit 31 reserved, since it's not clear how this would work for the scope bits like it would for things like the loop control bitfield or the memory operands bitfield.  Is this correct?

### DIFF
--- a/include/spirv/spir-v.xml
+++ b/include/spirv/spir-v.xml
@@ -293,8 +293,8 @@
     <ids type="ImageOperand" start="17" end="30" comment="Unreserved bits reservable for use by vendors"/>
     <ids type="ImageOperand" start="31" end="31" vendor="Khronos" comment="Reserved ImageOperand bit, not available to vendors"/>
 
-    <!-- SECTION: SPIR-V Scope Bit Reservations -->
-    <!-- Reserve ranges of bits in the "Scope"" bitfield.
+    <!-- SECTION: SPIR-V Scope Reservations -->
+    <!-- Reserve "Scope" values.
 
          Each vendor determines the use of values in their own ranges.
          Vendors are not required to disclose those uses.  If the use of a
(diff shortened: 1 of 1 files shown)


Tobski (approved):

bashbaug:
We observed in the January 31st telconference that the scope values aren't bits, so calling this a "scope bit" reservation is not correct and I need to update this PR.  I have marked this PR as a draft while I make the necessary changes.

Would we want to consider the scopes in their own "namespace" and reserve them via the XML file, similar to what is done in this PR aside from the "bit" naming?  Or, since the scopes are values, would we prefer to assign them from vendor-allocated enum blocks, in which case this PR may not actually be required?

I think I'll change this PR assuming the former, then if we prefer the latter we can simply close this PR without merging.

bashbaug:
OK, I've updated this PR so it refers to "scope values" and not "scope bits", but it seems like this shouldn't be necessary and we should just use an enum value from one of our previously allocated blocks for our new scope instead.  This could look a little weird since the KHR scope values will be small values (currently, 0 through 6), whereas the vendor scope values will be much larger (>4096), but this should be fine and we already have similar large discontinuities in other areas (e.g. Capabilities, Decorations, Builtins).

I'll keep this PR as a draft since I'm inclined to close it outright, without merging.

bashbaug:
Closing - this PR is not needed.  See discussion above.

PR #413, opened 2024-01-26
remove Kernel from Image Channel Order and Channel Data Type enums

bashbaug:
The Kernel requirement is already expressed transitively, since these enums are only returned by OpImageQueryOrder and OpImageQueryFormat, both which require the Kernel capability.

see: internal SPIR-V issue 773

### DIFF
--- a/include/spirv/unified1/spirv.core.grammar.json
+++ b/include/spirv/unified1/spirv.core.grammar.json
@@ -12198,121 +12198,101 @@
         {
           "enumerant" : "R",
           "value" : 0,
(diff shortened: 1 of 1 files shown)


bashbaug:
Update from the January 31st teleconference: may need some test updates in SPIR-V Tools.

alan-baker (approved):

dneto0 (approved):

bashbaug:
Merging offline, as discussed in the February 7th teleconference.

PR #418, opened 2024-02-13
cmake: Allow external control of test and install options

dneto0:
Test and install by default when the project is top level.

In any case, they can be controlled via CMake variables SPIRV_HEADERS_ENABLE_TEST and SPIRV_HEADERS_ENABLE_INSTALL

Rename the simple test target to spirv_headers_simple_test


bashbaug (approved):

bashbaug:
Merging as discussed in the February 14th teleconference.

PR #430, opened 2024-05-29
Reduce redundancy in utility functions

mmoult:
Commit `acfc25` introduced helpful utility functions to the C/C++ headers, but they have a lot of redundancy. These functions have long switch statements with trivial logic- ideal conditions for `#define`s. I did not see any style guide (perhaps the redundancy is unavoidable if `#define`s are forbidden), so I wrote this change.

Undoubtedly there are style considerations to do it the Khronos way. I figured I could submit this proposal and you can do with it what you will.

### DIFF
--- a/include/spirv/unified1/spirv.h
+++ b/include/spirv/unified1/spirv.h
@@ -1,20 +1,20 @@
 /*
 ** Copyright (c) 2014-2024 The Khronos Group Inc.
-** 
+**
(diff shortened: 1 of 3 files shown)


alan-baker:
These headers are all auto-generated. As it stands any future generation would override these changes. Take a look in the tools/buildHeaders directory for the code. I haven't thought too hard about the proposed changes themselves yet though.

PR #432, opened 2024-05-29
Enforce grammar naming conventions for aliases

### LINKED ISSUE #429: Enforce naming conventions for aliased tokens
Currently the grammar file has duplicate entries for when tokens from extensions are promoted to KHR or core. #426 added some utility methods for getting the names of enumerants. We should ensure to always follow a convention preferring the "highest" priority suffix:
1. No suffix (core)
2. KHR
3. EXT
4. Vendor

The best way to do this would be to modify the grammar to satisfy this ranking and enforce it via our build checks. 

This is related to #138, but easier to act on until we resolve the larger aliasing question.

alan-baker:
Fixes #429

* Enforce Core -> KHR -> EXT -> Vendor conventions for aliased names
* Update grammar to satisfy these conventions and regenerate headers

### DIFF
--- a/include/spirv/unified1/spirv.core.grammar.json
+++ b/include/spirv/unified1/spirv.core.grammar.json
@@ -5924,7 +5924,7 @@
       "version" : "None"
     },
     {
-      "opname" : "OpReportIntersectionNV",
+      "opname" : "OpReportIntersectionKHR",
       "class"  : "Reserved",
       "opcode" : 5334,
       "operands" : [
@@ -5938,7 +5938,7 @@
       "version" : "None"
     },
     {
-      "opname" : "OpReportIntersectionKHR",
+      "opname" : "OpReportIntersectionNV",
       "class"  : "Reserved",
       "opcode" : 5334,
       "operands" : [
@@ -6055,7 +6055,7 @@
       "version" : "None"
     },
     {
-      "opname" : "OpTypeAccelerationStructureNV",
+      "opname" : "OpTypeAccelerationStructureKHR",
       "class"  : "Type-Declaration",
       "opcode" : 5341,
       "operands" : [
@@ -6066,7 +6066,7 @@
       "version" : "None"
     },
     {
-      "opname" : "OpTypeAccelerationStructureKHR",
+      "opname" : "OpTypeAccelerationStructureNV",
       "class"  : "Type-Declaration",
       "opcode" : 5341,
       "operands" : [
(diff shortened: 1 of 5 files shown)


dneto0 (approved):
LGTM. This will cause disassembly to change.  So it may require SPIRV-Tools updates.

include/spirv/unified1/spirv.core.grammar.json
```diff
@@ -5938,7 +5938,7 @@
       "version" : "None"
     },
     {
-      "opname" : "OpReportIntersectionKHR",
+      "opname" : "OpReportIntersectionNV",
```
kpet, line 5941: Not for this change but this makes me wonder whether we shouldn't introduce a way of encoding aliases that does not require duplicating whole definitions. The first idea that comes to mind is something like:

```
{
  "opname": "OpReportIntersectionKHR",
  "aliases": ["OpReportIntersectionNV"]
}
```

Is there a reason we wouldn't want to do something like this?

tools/buildHeaders/jsonToSpirv.cpp
```diff
@@ -471,6 +471,36 @@ unsigned int NumberStringToBit(const std::string& str)
...
+// * Core
+// * KHR
+// * EXT
+// * Vendor (no preference between vendors)
+//
+// Returns true if the order is valid.
+bool SuffixComparison(const std::string& prev, bool prevCore,
+                      const std::string& cur, bool curCore)
+{
+  if (prev == cur) return true;
```
kpet, line 485: IIUC, this condition is only hit when encountering two instructions with the same name. Shouldn't we error out instead? Do we reject duplicates somewhere else?

kpet (commented):
Thanks for doing this!

include/spirv/unified1/spirv.core.grammar.json
alan-baker: That's the topic of #138

tools/buildHeaders/jsonToSpirv.cpp
alan-baker: We didn't reject duplicates previously, so I made that change.

include/spirv/unified1/spirv.core.grammar.json
alan-baker: My opinion is that we should do something along those lines, but it will take longer to nail down how we want it to work (e.g. in conjunction with the spec). So for now this is desirable.

kpet: Yeah, it's a much bigger change. I'm not suggesting we try to do this here and now, just checking that I hadn't missed a reason we couldn't (or didn't want to) go there.

tools/buildHeaders/jsonToSpirv.cpp
kpet: Good to do IMO, it buys us an extra check for very little work. Thanks!

kpet (approved):

bashbaug:
Merging as discussed in the June 5th teleconference.

PR #443, opened 2024-08-13
Request a vendor id for Kongruent (my new shader compiler)

RobDangerous:
Hello!

I built a new shader-(cross-)compiler and would like to request a vendor id. It creates SPIR-V directly (among other output options).

Best regards,
Robert


Naghasan (approved):

bashbaug:
Merging as discussed in the August 14th teleconference.

PR #447, opened 2024-09-11
Add "aliases" fields to the grammar and remove duplicated

Naghasan:
Promotion of extensions lead to duplicated entries. The patch introduce a new entry "aliases" to the grammar. It is a list of strings, where each string is an alternative name for the token, and applies on instructions and operand kind's enumerants. The grammar is updated to use the fields and all duplicated were removed.

The header generator now processes the aliases when generating the headers to account for the change. This ensure the generated headers doesn't change.

Note: this change is breaking for tools processing the grammar directly, tools relying on the generated header aren't affected.

The PR supersedes https://github.com/KhronosGroup/SPIRV-Headers/pull/138 and https://github.com/KhronosGroup/SPIRV-Headers/pull/139


alan-baker:
https://github.com/KhronosGroup/SPIRV-Tools/pull/5799 handles the tooling side of this change.

dneto0 (approved):
LGTM.
Thank you, this is so much better.  :-)

alan-baker:
2024-09-25 SPIR WG telecon: Set a feedback date of October 9 prior to accepting this change.

bashbaug:
Merging as discussed in the October 9th teleconference.

spencer-lunarg:
Just hit this change now rebasing, this is awesome! ... time to throw away all the old code that use to do this manually

PR #450, opened 2024-09-24
Add SPV_EXT_arithmetic_fence

MrSidims:
Spec:
https://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/EXT/SPV_EXT_arithmetic_fence.html


bashbaug (approved):
LGTM.

Having both the INTEL names and the EXT names is a bit unfortunate, especially since SPV_INTEL_arithmetic_fence was never formally published.  Do we really need both?

Naghasan (approved):

bashbaug:
Merging as discussed in the October 2nd teleconference.

PR #455, opened 2024-10-15
Remove extension from OpDecorateId

Naghasan:
OpDecorateId was introduced by SPIR-V 1.2 not by SPV_GOOGLE_hlsl_functionality1

### DIFF
--- a/include/spirv/unified1/spirv.core.grammar.json
+++ b/include/spirv/unified1/spirv.core.grammar.json
@@ -3863,7 +3863,6 @@
         { "kind" : "IdRef", "name" : "'Target'" },
         { "kind" : "Decoration" }
       ],
-      "extensions" : [ "SPV_GOOGLE_hlsl_functionality1" ],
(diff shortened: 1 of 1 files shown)


bashbaug:
It might be both SPIR-V 1.2 and SPV_GOOGLE_hlsl_functionality1.  See:
https://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/GOOGLE/SPV_GOOGLE_hlsl_functionality1.html

> In the OpDecorateId instruction’s Capability box, remove the statement "Missing before version 1.2", to allow this instruction to be used with this extension. Also, in the description of the OpDecorateId instruction, change "All such <id> Extra Operands must be constant instructions." to "All such <id> Extra Operands must be constant instructions or the result <id> of OpVariable instructions.

Naghasan:
true the extension enables it outside 1.2, I made the shortcut of "adds" rather than "enables"

PR #456, opened 2024-10-16
Add scope resolution operator in spirv.hpp11

Christian-Conti:
The file spirv.hpp11 was missing of the scope resolution operator :: in the cases insde the funcions to separate the namespace from the member

PR #461, opened 2024-10-25
add basic utility code testing for cpp, cpp11, and c

bashbaug:
fixes #457

These tests will fail until the issue described in #463 is resolved.

PR #463, opened 2024-10-29
Fix on header generator for c++11, regenerated products

Naghasan:
Rework from https://github.com/KhronosGroup/SPIRV-Headers/pull/456 (new PR to unlock the CLA issue)


include/spirv/unified1/spirv.hpp11
```diff
@@ -3025,1075 +3025,1075 @@ inline void HasResultAndType(Op opcode, bool *hasResult, bool *hasResultType) {
...
-    case Dim2D: return "2D";
-    case Dim3D: return "3D";
-    case DimCube: return "Cube";
-    case DimRect: return "Rect";
-    case DimBuffer: return "Buffer";
-    case DimSubpassData: return "SubpassData";
-    case DimTileImageDataEXT: return "TileImageDataEXT";
+    case Dim::1D: return "1D";
+    case Dim::2D: return "2D";
+    case Dim::3D: return "3D";
```
bashbaug, line 3225: The names of these enums should be e.g. `Dim::Dim1D` instead of `Dim::1D`, to avoid identifiers beginning with numbers.

bashbaug (changes requested):
I found this issue by testing these changes with the tests from #461, so the tests are working.

https://github.com/bashbaug/SPIRV-Headers/tree/add-utility-code-testing-test

include/spirv/unified1/spirv.hpp11
Naghasan: as you said, we know the test are working. I fixed it and tested locally, so should be good now

bashbaug (approved):

Issue #460, opened 2024-10-23
Add provisional information in the grammar


spencer-lunarg:
The changes in https://github.com/KhronosGroup/SPIRV-Headers/pull/452 broke backward compatibility, which is find because it provisional, the issue is in Vulkan we have a 

```c
#ifdef VK_ENABLE_BETA_EXTENSIONS
#include "vulkan_beta.h"
#endif
```

that wraps https://github.com/KhronosGroup/Vulkan-Headers/blob/main/include/vulkan/vulkan_beta.h such that tools/apps can build with/without these provisional extensions... SPIR-V has no concept and there is no automated way in the grammar (and more importantly in the headers that are breaking backwards compatibility for everyone) to know what is provisional

spencer-lunarg:
WG - 10/30/24 

agreed we should do this, I will add a field to the grammar and use the Validation Layers to tests it all and come back with a full solution

PR #464, opened 2024-11-13
Add provisional key to grammar

spencer-lunarg:
closes https://github.com/KhronosGroup/SPIRV-Headers/issues/460

The idea is like `VK_ENABLE_BETA_EXTENSIONS` the consumer of the headers can now consume the SPIR-V if they desire ( by adding something like `add_compile_definitions(SPV_ENABLE_BETA_EXTENSIONS)` to their build)

Open to suggestions on other ways to go about this

### DIFF
--- a/include/spirv/unified1/spirv.core.grammar.json
+++ b/include/spirv/unified1/spirv.core.grammar.json
@@ -5373,6 +5373,7 @@
         { "kind" : "IdRef", "name": "'Node Index'" }
       ],
       "capabilities" : [ "ShaderEnqueueAMDX" ],
+      "provisional" : true,
       "version" : "None"
     },
     {
@@ -5383,6 +5384,7 @@
         { "kind" : "IdRef", "name": "'Payload Array'" }
       ],
       "capabilities" : [ "ShaderEnqueueAMDX" ],
+      "provisional" : true,
       "version" : "None"
     },
     {
@@ -5394,6 +5396,7 @@
         { "kind" : "IdRef", "name": "'Payload Type'" }
       ],
       "capabilities" : [ "ShaderEnqueueAMDX" ],
+      "provisional" : true,
       "version" : "None"
     },
     {
@@ -5406,6 +5409,7 @@
         { "kind" : "IdRef", "name": "'Payload'" }
       ],
       "capabilities" : [ "ShaderEnqueueAMDX" ],
+      "provisional" : true,
       "version" : "None"
     },
     {
@@ -5418,6 +5422,7 @@
         { "kind" : "IdRef", "name": "'Payload Array'" }
       ],
       "capabilities" : [ "ShaderEnqueueAMDX" ],
+      "provisional" : true,
       "version" : "None"
     },
     {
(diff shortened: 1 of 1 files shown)


tools/buildHeaders/header.cpp
```diff
@@ -123,6 +124,8 @@ namespace {
                                enumStyle_t style) const {
             return enumFmt(s, v, style, true);
         }
+        virtual std::string enumProvisionalStart(const std::string& name) const { return ""; };
+        virtual std::string enumProvisionalEnd(const std::string& name)  const { return ""; };
```
Naghasan, line 128:
```suggestion
        virtual std::string guardIfProvisionalStart(const std::string& name) const { return ""; };
        virtual std::string guardIfProvisionalEnd(const std::string& name)  const { return ""; };
```

just to convey better the function checks before emitting, otherwise it would be better to hoist the check out of the function

spencer-lunarg:
Thinking about this for a night, I realize maybe the `#ifdef` in the header isn't needed

To my knowledge what is breaking people is not so much having some `AMDX` extension change in the header, but actually when people are parsing the grammar themselves.

Maybe the simpler first step is to just add the `provisional` to the grammar

Naghasan:
> Maybe the simpler first step is to just add the provisional to the grammar

Make sense and I agree. We only generated enums and utils against this enum. So if a user of the generated headers needs to guard a provisional spec they'll need one on their side anyway.

Naghasan (approved):
LGTM, Thanks Alan

bashbaug:
We discussed this PR in the November 13th teleconference.  Everything looks reasonable, but it'd be nice to have an example using these provisional additions to the grammar before merging to increase confidence it has everything that we need and so we don't make changes again.

@spencer-lunarg do you plan to use the provisional key in the validation layers?  If not, maybe we should resurrect the header changes?

dneto0:
This docs for the JSON grammar should be updated too.
See https://registry.khronos.org/SPIR-V/specs/unified1/MachineReadableGrammar.html 

Although I'm having trouble finding the source for that (!)

dneto0:
What exactly does "provisional" mean, and what is going to act on it?  I feel it's a little loose.

> This docs for the JSON grammar should be updated too. See https://registry.khronos.org/SPIR-V/specs/unified1/MachineReadableGrammar.html
> 
> Although I'm having trouble finding the source for that (!)

oh, it's in the Khronos-internal SPIR-V spec repo under specs/MachineReadableGrammar.asciidoc

bashbaug:
> What exactly does "provisional" mean, and what is going to act on it?

Good question.  The name "provisional" comes from "provisionally ratified", but I think it's being used to mean "hey this isn't final, so it might change or go away".

If we update the header generation scripts to act on it we could add preprocessor guards around the "provisional" features, to provide a way for compilers or tools to opt-out of the features that are likely to break compatibility for example (see #460).

Maybe we should use a different name instead?

> This docs for the JSON grammar should be updated too.

Yep, agreed, and @Naghasan already filed a Khronos issue so this doesn't slip through the cracks.  Thanks for caling this out though!

spencer-lunarg:
Created https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8870 as a showcase of how Validation Layers would use this new grammar

PR #468, opened 2024-12-19
Fix OpAsmTargetINTEL operand

jdnk:
The `OpAsmTargetINTEL` instruction had an extra Result Type operand which is not specified in the extension https://github.com/intel/llvm/blob/sycl/sycl/doc/design/spirv-extensions/SPV_INTEL_inline_assembly.asciidoc. This causes errors in tools relying on the headers.

Small example of the failure:
```c
// foo.c
void foo (const float * restrict a, const float * restrict b, float * restrict c) {
    __asm__ ("nop");
}
```

Compile to `.spt` via `llc`:
```bash
clang -S -emit-llvm -O3 foo.c -o foo.ll
llc -O3 -mtriple=spirv64-unknown-unknown --spirv-ext=+SPV_INTEL_inline_assembly foo.ll -o foo.spt
```

The output contains something like `%12 = OpAsmTargetINTEL "spirv64-unknown-unknown"`, without any result type. Then, running `spirv-as foo.spt -o foo.spv` fails because return type is expected.

Tagging @mmerecki who worked on the extension.

### DIFF
--- a/include/spirv/unified1/spirv.core.grammar.json
+++ b/include/spirv/unified1/spirv.core.grammar.json
@@ -6986,7 +6986,6 @@
       "class"  : "@exclude",
       "opcode" : 5609,
       "operands" : [
-        { "kind" : "IdResultType" },
         { "kind" : "IdResult" },
         { "kind" : "LiteralString", "name" : "'Asm target'" }
       ],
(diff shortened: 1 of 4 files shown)


Naghasan (approved):

mmerecki (approved):

bashbaug:
Merging as discussed in the January 8th teleconference.

PR #475, opened 2025-01-21
Use modern asciidoc markup in JSON annotations, instead of compat-mode

oddhack:
This is part of ongoing work to integrate the SPIR-V specs into the Vulkan Antora docs site build. While the internal spec tree builds from an internal (slightly laggy) mirror of this repo, @Naghasan recommended I submit the PR here instead. We might want to add a CI check for the old-style markup for a while, since that will be easy to overlook especially in currently open PRs. I can propose that here or separately (would not amount to more than grepping for

```
"'.*'"
```

in include/spirv/unified1/*.json, and failing the CI check if matched.

As far as I can tell, this makes no difference at all to the outputs when generating spec HTML from the default branch of the internal spec tree (as it should not).

### DIFF
--- a/include/spirv/unified1/extinst.debuginfo.grammar.json
+++ b/include/spirv/unified1/extinst.debuginfo.grammar.json
@@ -35,253 +35,253 @@
       "opname" : "DebugCompilationUnit",
       "opcode" : 1,
       "operands" : [
-        { "kind" : "IdRef", "name" : "'Source'" },
-        { "kind" : "LiteralInteger", "name" : "'Version'" },
-        { "kind" : "LiteralInteger", "name" : "'DWARF Version'" }
+        { "kind" : "IdRef", "name" : "Source" },
+        { "kind" : "LiteralInteger", "name" : "Version" },
+        { "kind" : "LiteralInteger", "name" : "DWARF Version" }
       ]
     },
     {
       "opname" : "DebugTypeBasic",
       "opcode" : 2,
       "operands" : [
-        { "kind" : "IdRef", "name" : "'Name'" },
-        { "kind" : "IdRef", "name" : "'Size'" },
-        { "kind" : "DebugBaseTypeAttributeEncoding", "name" : "'Encoding'" }
+        { "kind" : "IdRef", "name" : "Name" },
+        { "kind" : "IdRef", "name" : "Size" },
+        { "kind" : "DebugBaseTypeAttributeEncoding", "name" : "Encoding" }
       ]
     },
     {
       "opname" : "DebugTypePointer",
       "opcode" : 3,
       "operands" : [
-        { "kind" : "IdRef", "name" : "'Base Type'" },
-        { "kind" : "StorageClass", "name" : "'Storage Class'" },
-        { "kind" : "DebugInfoFlags", "name" : "'Literal Flags'" }
+        { "kind" : "IdRef", "name" : "Base Type" },
+        { "kind" : "StorageClass", "name" : "Storage Class" },
+        { "kind" : "DebugInfoFlags", "name" : "Literal Flags" }
       ]
     },
     {
       "opname" : "DebugTypeQualifier",
       "opcode" : 4,
       "operands" : [
-        { "kind" : "IdRef", "name" : "'Base Type'" },
-        { "kind" : "DebugTypeQualifier", "name" : "'Type Qualifier'" }
+        { "kind" : "IdRef", "name" : "Base Type" },
+        { "kind" : "DebugTypeQualifier", "name" : "Type Qualifier" }
       ]
     },
     {
       "opname" : "DebugTypeArray",
       "opcode" : 5,
       "operands" : [
-        { "kind" : "IdRef", "name" : "'Base Type'" },
-        { "kind" : "IdRef", "name" : "'Component Counts'", "quantifier" : "*" }
+        { "kind" : "IdRef", "name" : "Base Type" },
+        { "kind" : "IdRef", "name" : "Component Counts", "quantifier" : "*" }
       ]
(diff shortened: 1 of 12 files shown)


bashbaug (commented):
I didn't look at every one of the changes, but this all looks fairly mechanical and I didn't spot any obvious issues.  Was there anything that wasn't mechanical that is worth calling out and reviewing in more detail?

For the purposes of this repo, none of these names get folded into the headers, so this all seems just fine.  I think it will affect the spec though, even if it's just at the asciidoctor source level, so I'd like to wait for @Naghasan 's signoff before merging.

Longer term, I wonder if some of all of this asciidoc markup can be folded into the spec generator vs. including it in the SPIR-V grammar file itself.  This way, the grammar file wouldn't need to change if we decided to adopt some other syntax, or if somebody wants to do something based on the operand names in the grammar using a non-asciidoc toolchain.

oddhack:
> I didn't look at every one of the changes, but this all looks fairly mechanical and I didn't spot any obvious issues. Was there anything that wasn't mechanical that is worth calling out and reviewing in more detail?

No.

> For the purposes of this repo, none of these names get folded into the headers, so this all seems just fine. I think it will affect the spec though, even if it's just at the asciidoctor source level, so I'd like to wait for @Naghasan 's signoff before merging.


I generated HTML from main and from this branch and they are identical, AFAICT, but I did want @Naghasan to double-check that.

> Longer term, I wonder if some of all of this asciidoc markup can be folded into the spec generator vs. including it in the SPIR-V grammar file itself. This way, the grammar file wouldn't need to change if we decided to adopt some other syntax, or if somebody wants to do something based on the operand names in the grammar using a non-asciidoc toolchain.

I did notice that not all the 'name' fields are simple operand names, some are lists like 'op1, op2, ... opN' which might complicate that slightly. If we know that's the only irregularity then folding it into the generator script sounds good.

dneto0:
The SPIRV-Tools build also consumes the .json files.  I tried this PR with the latest SPIRV-Tools and shaderc builds, and all the tests pass. So this PR LGTM from my perspective.

cc: @alan-baker

dneto0:
I posted #477 which adds the CI grep check that @oddhack  suggested.  Seems to work, as shown by #478

oddhack:
> I posted #477 which adds the CI grep check that @oddhack suggested. Seems to work, as shown by #478

LGTM. It looks like #477 started from this branch - would it make sense to close this and just merge #477 instead?

Naghasan:
@oddhack Thanks for the PR helps moving the specs structure in the right direction.

> Longer term, I wonder if some of all of this asciidoc markup can be folded into the spec generator vs. including it in the SPIR-V grammar file itself. This way, the grammar file wouldn't need to change if we decided to adopt some other syntax, or if somebody wants to do something based on the operand names in the grammar using a non-asciidoc toolchain.

We discussed this during teleconferences, I think it is the right opportunity to remove these annotations. The spec generator should be the one to decide how to format the field and it potentially allows tool to freely use the field. Another motivation is the field may have mistake causing rendering issues in the specs (like here https://github.com/KhronosGroup/SPIRV-Headers/pull/485), I don't think contributors should have to care about this.

For the list like names, entries like `'Member 0 type', +\n'member 1 type', +\n...` should simply become `Member type` and the spec generator will then expand it to `_Member type 0_, +\n_Member type 1_, +\n...` thanks to the quantifier.

oddhack:
> We discussed this during teleconferences, I think it is the right opportunity to remove these annotations. The spec generator should be the one to decide how to format the field and it potentially allows tool to freely use the field. Another motivation is the field may have mistake causing rendering issues in the specs (like here #485), I don't think contributors should have to care about this.

For the 'name' entries in the JSON files that seems fine, but I notice that there's a lot of hardcoded spec markup in the generator cpp files (in the spec source repository, not here) as well, and that would be a lot harder to inject markup into were it removed, since that's essentially free-form spec language.

My inclination for what I'm doing here is to strip out the markup from the JSON 'name' fields and re-insert that via generators as may be required, but leave the other source files alone. Will that work for you @Naghasan ?

Naghasan:
> My inclination for what I'm doing here is to strip out the markup from the JSON 'name' fields and re-insert that via generators as may be required, but leave the other source files alone.

Sounds good to me to simply remove this.

> the other source files alone.

I'm not too sure to understand which one you are referring too

oddhack:
> > My inclination for what I'm doing here is to strip out the markup from the JSON 'name' fields and re-insert that via generators as may be required, but leave the other source files alone.
> 
> Sounds good to me to simply remove this.

Wilco.
 
> > the other source files alone.
> 
> I'm not too sure to understand which one you are referring too

I'm referring to all the asciidoc markup in the generator .cpp files themselves.

BTW, I'm also running into a build issue I hope you can help with. I'd prefer to do the PR here rather than against the internal SPIRV-Headers mirror, since this is the most up to date. But when I attempt to build the SPIR-V spec repo from gitlab against this version of the header repo, I'm seeing some error messages which appear related to recently added Intel enum information. TBH I don't really understand how the generator works so am poking at it somewhat blindly and just comparing the HTML outputs. 

Hopefully one end result of this will be that when we get the Antora conversion proceeding and get a mirror of the spec repo onto gitlab, CI can be setup to catch such minor inconsistencies between the two repos.

Naghasan:
> I'd prefer to do the PR here rather than against the internal SPIRV-Headers mirror, since this is the most up to date.

I synced the internal, but here is preferable IMO.

> TBH I don't really understand how the generator works so am poking at it somewhat blindly and just comparing the HTML outputs.

Lets sync offline.

Naghasan:
Merging as agreed during call

PR #477, opened 2025-01-22
Ci legacy asciidoc

dneto0:
Adds a CI check for #475


Naghasan:
Merging as agreed during call

PR #479, opened 2025-01-24
updates AtomicStorageOps to depend on the AtomicStorage capability

### LINKED ISSUE #473: AtomicStorageOps is missing dependency on AtomicStorage
According to the [SPV_KHR_shader_atomic_counter_ops](https://github.khronos.org/SPIRV-Registry/extensions/KHR/SPV_KHR_shader_atomic_counter_ops.html) extension, AtomicStorageOps should depend on / implicitly declare AtomicStorage, but this dependency isn't described in the grammar file:

https://github.com/KhronosGroup/SPIRV-Headers/blob/2b2e05e088841c63c0b6fd4c9fb380d8688738d3/include/spirv/unified1/spirv.core.grammar.json#L15712-L15717

Since the extension spec refers to the dependency in several places, I think the grammar file should be updated and not the extension text, but either could be modified to fix the inconsistency.

bashbaug:
fixes #473 

Updates AtomicStorageOps to depend on / implicitly declare the AtomicStorage capability, fixing an inconsistency between the [spec](https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/KHR/SPV_KHR_shader_atomic_counter_ops.asciidoc#modifications-to-the-spir-v-specification-version-11) and the SPIR-V grammar.

### DIFF
--- a/include/spirv/unified1/spirv.core.grammar.json
+++ b/include/spirv/unified1/spirv.core.grammar.json
@@ -15712,6 +15712,7 @@
         {
           "enumerant" : "AtomicStorageOps",
           "value" : 4445,
+          "capabilities" : [ "AtomicStorage" ],
           "extensions" : [ "SPV_KHR_shader_atomic_counter_ops" ],
           "version" : "None"
         },


Naghasan (approved):
LGTM, the fix shouldn't be really an issue in practice.

PR #481, opened 2025-01-28
updates IntegerFunctions2INTEL to remove Shader capability dependency

### LINKED ISSUE #480: IntegerFunctions2INTEL has extra dependency on Shader
According to the [SPV_INTEL_shader_integer_functions2](https://github.khronos.org/SPIRV-Registry/extensions/INTEL/SPV_INTEL_shader_integer_functions2.html) specification, the IntegerFunctions2INTEL capability added by this extension should not depend on any other capabilities, but in the SPIR-V grammar file there is an implicit dependency on Shader:

https://github.com/KhronosGroup/SPIRV-Headers/blob/2b2e05e088841c63c0b6fd4c9fb380d8688738d3/include/spirv/unified1/spirv.core.grammar.json#L16315-L16321

We should fix this inconsistency.  Because I don't see anything in the descriptions of these instructions that would make them Shader specific I would prefer to update the grammar file vs. updating the spec, but either is possible.

CC: @ianromanick

bashbaug:
fixes #480 

Updates the IntegerFunctions2INTEL capability to remove the Shader capability dependency, because no dependency is described in the SPV_INTEL_shader_integer_functions2 [spec](https://github.khronos.org/SPIRV-Registry/extensions/INTEL/SPV_INTEL_shader_integer_functions2.html).

### DIFF
--- a/include/spirv/unified1/spirv.core.grammar.json
+++ b/include/spirv/unified1/spirv.core.grammar.json
@@ -16315,7 +16315,6 @@
         {
           "enumerant" : "IntegerFunctions2INTEL",
           "value" : 5584,
-          "capabilities" : [ "Shader" ],
           "extensions" : [ "SPV_INTEL_shader_integer_functions2" ],
           "version" : "None"
         },


Naghasan (approved):

bashbaug:
Merging as discussed in the January 29th teleconference.

PR #496, opened 2025-03-07
Update Slang tool entry.

csyonghe:
Slang is now under Khronos, so we need to update the vendor field to reflect the change.


alan-baker (approved):

bmillsNV (approved):

bashbaug:
Merging as discussed in the March 12th teleconference.

PR #498, opened 2025-03-12
Fix json capabilities for RayTracingOpacityMicromapEXT

dgkoch:
Fix Issue #319

### DIFF
--- a/include/spirv/unified1/spirv.core.grammar.json
+++ b/include/spirv/unified1/spirv.core.grammar.json
@@ -16587,7 +16587,7 @@
         {
           "enumerant" : "RayTracingOpacityMicromapEXT",
           "value" : 5381,
-          "capabilities" : [ "RayQueryKHR","RayTracingKHR" ],
+          "capabilities" : [ "Shader" ],
           "extensions" : [ "SPV_EXT_opacity_micromap" ],
           "version" : "None"
         },


Naghasan (approved):

kimwooyoung:
I feel like it makes less sense to not make the capability imply the RayQueryKHR/RayTracingKHR capabilities as the flag is useless unless a shader that uses it has a traceRay call.

dgkoch:
> I feel like it makes less sense to not make the capability imply the RayQueryKHR/RayTracingKHR capabilities as the flag is useless unless a shader that uses it has a traceRay call.

Well this matches the spec, and as this was previously written it would require *both* RayQuery and RayTracing which isn't correct either.  Presumably the Query or Trace capability dependency should be just as well satisfied by the Trace or query instructions themselves, so it all seems rather academic.

bashbaug (approved):
LGTM.

Noting for completeness: this change could make a previously valid SPIR-V module invalid, if it was relying on the **RayTracingOpacityMicromapEXT** capability to implicitly declare **RayQueryKHR** and/or **RayTracingKHR** and did not declare these capabilities explicitly.  Do we think this will ever occur in practice?

kimwooyoung:
> Noting for completeness: this change could make a previously valid SPIR-V module invalid, if it was relying on the RayTracingOpacityMicromapEXT capability to implicitly declare RayQueryKHR and/or RayTracingKHR and did not declare these capabilities explicitly. Do we think this will ever occur in practice?

I think there is a remote possibility. The initial OMM commit to glslang did not even generate the RayTracingOpacityMicromapEXT capability, so the issue won't occur for SPV shaders generated from glslang, though.

> Well this matches the spec, and as this was previously written it would require both RayQuery and RayTracing which isn't correct either. Presumably the Query or Trace capability dependency should be just as well satisfied by the Trace or query instructions themselves, so it all seems rather academic.

The json file is used for automated header generation. The actual capability generation is done by compilers such as glslang, and spirv-val implements the validation, so I believe what is in the json file does not necessarily "requires" both RayQuery and RayTracing. If we are to fix the spec, we could explicitly state that RayTracing for ray pipeline shaders and RayQuery for the rest are implicitly declared.

dgkoch:
> Noting for completeness: this change could make a previously valid SPIR-V module invalid, if it was relying on the RayTracingOpacityMicromapEXT capability to implicitly declare RayQueryKHR and/or RayTracingKHR and did not declare these capabilities explicitly. Do we think this will ever occur in practice?

This extension only adds a single ray flag. If the shader wasn't already using something else that requires RayQuery or RayTracing, having the RayTracingOpacityMicromapEXT capability declared won't make any difference, as it's not actually using the functionality. 

> If we are to fix the spec, we could explicitly state that RayTracing for ray pipeline shaders and RayQuery for the rest are implicitly declared.

Not sure what you are proposing to "fix" in the spec. Those capabilities are already required for any of the instructions which could consume the new ray flag.

bashbaug:
Merging as discussed in the March 19th teleconference.

PR #502, opened 2025-03-18
Add headers for SPV_KHR_bfloat16

stu-s:


dneto0 (approved):
This matches the extension text as rendered internally to Khronos
from
https://gitlab.khronos.org/spirv/spirv-extensions/-/merge_requests/243

PR #503, opened 2025-03-20
Fix OpFPGARegINTEL grammar

### LINKED ISSUE #497: Mismatch in Operand Count for OpFPGARegINTEL in SPIR-V Headers
The JSON grammar definition for OpFPGARegINTEL in the SPIR-V headers appears to include an extra operand that is not specified by the SPIR-V specification. According to the spirv-specifications the operands of this instruction are:
```
Result Type
Result
Input
```
Specification:
[](https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/INTEL/SPV_INTEL_fpga_reg.asciidoc)
However, the current JSON grammar for OpFPGARegINTEL is defined as follows:

```
{
  "opname" : "OpFPGARegINTEL",
  "class"  : "Reserved",
  "opcode" : 5949,
  "operands" : [
    { "kind" : "IdResultType" },
    { "kind" : "IdResult" },
    { "kind" : "IdRef", "name" : "'Result'" },
    { "kind" : "IdRef", "name" : "'Input'" }
  ],
  "capabilities" : [ "FPGARegINTEL" ],
  "extensions" : [ "SPV_INTEL_fpga_reg" ],
  "version" : "None"
}
```
Here an additional operand Result IdRef is present. This is causing validation issues. Tests are passing in bidirectional translator. However on trying to use spirv-val, it shows error:
`error: 242: End of input reached while decoding OpFPGARegINTEL starting at word 1012: expected more operands after 4 words.`

Is the issue in SPIRV Headers itself? If yes I would like to correct this and contribute.

Naghasan:
Fix #497

### DIFF
--- a/include/spirv/unified1/spirv.core.grammar.json
+++ b/include/spirv/unified1/spirv.core.grammar.json
@@ -9999,7 +9999,6 @@
       "operands" : [
         { "kind" : "IdResultType" },
         { "kind" : "IdResult" },
-        { "kind" : "IdRef", "name" : "'Result'" },
         { "kind" : "IdRef", "name" : "'Input'" }
       ],
       "capabilities" : [ "FPGARegINTEL" ],


Naghasan:
Tested locally: without this change, spirv-dis cannot process the module created by this test https://github.com/intel/llvm/blob/sycl/llvm-spirv/test/extensions/INTEL/SPV_INTEL_fpga_reg/IntelFPGAReg.ll

bashbaug (approved):
LGTM, changes match the [spec](https://github.khronos.org/SPIRV-Registry/extensions/INTEL/SPV_INTEL_fpga_reg.html).

bashbaug:
Merging as discussed in the March 26th teleconference.

PR #508, opened 2025-04-04
Add DisplacementMicromapNV to the capabilities enabling OpTypeAcceler…

rj123-nv:
…ationStructureKHR

Spirv modules declaring this capability should be able to use acceleration structures without validation errors.
Spec here: https://github.khronos.org/SPIRV-Registry/extensions/NV/SPV_NV_displacement_micromap.html

### DIFF
--- a/include/spirv/unified1/spirv.core.grammar.json
+++ b/include/spirv/unified1/spirv.core.grammar.json
@@ -6300,8 +6300,8 @@
       "operands" : [
         { "kind" : "IdResult" }
       ],
-      "capabilities" : [ "RayTracingNV" , "RayTracingKHR", "RayQueryKHR" ],
-      "extensions" : [ "SPV_NV_ray_tracing" , "SPV_KHR_ray_tracing", "SPV_KHR_ray_query" ],
+      "capabilities" : [ "RayTracingNV" , "RayTracingKHR", "RayQueryKHR", "DisplacementMicromapNV" ],
+      "extensions" : [ "SPV_NV_ray_tracing" , "SPV_KHR_ray_tracing", "SPV_KHR_ray_query", "SPV_NV_displacement_micromap" ],
       "version" : "None"
     },
     {


dgkoch (approved):
Does what the spec says. 

In retrospect, I think it should have been an indirect dependency to RayTracingKHR or RayQueryKHR for the acceleration structure, but DMM has been deprecated so I don't think there is any interested in attempting to change it at this point.

include/spirv/unified1/spirv.core.grammar.json
```diff
@@ -6300,8 +6300,8 @@
       "operands" : [
         { "kind" : "IdResult" }
       ],
-      "capabilities" : [ "RayTracingNV" , "RayTracingKHR", "RayQueryKHR" ],
-      "extensions" : [ "SPV_NV_ray_tracing" , "SPV_KHR_ray_tracing", "SPV_KHR_ray_query" ],
+      "capabilities" : [ "RayTracingNV" , "RayTracingKHR", "RayQueryKHR", "DisplacementMicromapNV" ],
+      "extensions" : [ "SPV_NV_ray_tracing" , "SPV_KHR_ray_tracing", "SPV_KHR_ray_query", "SPV_NV_displacement_micromap" ],
```
bashbaug, line 6304: Checking: Do we really need to / want to add to the list of extensions for this instruction?

The instruction is already in the "allow list" for "doubly-enabled" instructions, so it's not causing any of the automated checks to fail:

https://github.com/KhronosGroup/SPIRV-Headers/blob/8e82b7cfeca98baae9a01a53511483da7194f854/tools/buildHeaders/jsonToSpirv.cpp#L100

I'm not sure if it'd be better to omit this extension though, following best practices for when to explicitly list the enabling extension, or whether it'd be better to list the extension for consistency.

dgkoch: (repeating what I said on the call): 
It is possible to use some of the instructions in the DMM extension with OpTypeAccelerationStructure directly in a mesh shader without using RT or RQ query, so I think this change is "correct". That said, we've stopped supporting this provisional extension, so I'm not sure this is going to matter in the real world any longer.

PR #511, opened 2025-04-20
Add Bazel bzlmod support

abrisco:
This change adds Bazel bzlmod support so this repository can be added to the [Bazel Central Registry](https://registry.bazel.build/)

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -1,0 +1,8 @@
+"""SPIRV-Headers"""
+
+module(
+    name = "spirv_headers",
+    version = "0.0.0",
+)
+
+bazel_dep(name = "rules_cc", version = "0.1.1")
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 build
 out
 .DS_Store
+bazel-*
+MODULE.bazel.lock
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -1,3 +1,5 @@
+load("@rules_cc//cc:cc_library.bzl", "cc_library")
+
 package(
     default_visibility = ["//visibility:public"],
 )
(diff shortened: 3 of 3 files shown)


MODULE.bazel
```diff
@@ -0,0 +1,8 @@
+"""SPIRV-Headers"""
+
+module(
+    name = "spirv_headers",
+    version = "0.0.0",
```
abrisco, line 5: Please note that this value is just a placeholder.

dnovillo: Where do you expect to use this? Should we have a more significant value for `version`?

dnovillo (changes requested):
LGTM.  One question on the `version` value.

MODULE.bazel
abrisco: This will get used by the Bazel Central Registry (BCR) where this version will be overwritten with something to uniquely identify the version. The guidance for repositories that don't do semver releases is to use `0.0.0-YYYMMDD-{commit_hash}`. Bazel consumers will then specify the same version in their `MODULE.bazel` file. If there is an actual version number for these headers then that could be added here but I didn't want to increase the maintenance burden of these files and figured it would be fine to correctly reflect the version whenever someone wants a new release published to BCR.

Does that answer your question?

dnovillo: Yup. Thanks. I was not too familiar with how the BCR works.

dnovillo (approved):

dnovillo:
Are you able to merge it yourself? If not, I'll merge it for you.

abrisco:
> Are you able to merge it yourself? If not, I'll merge it for you.

I don't think I'm able to merge since I'm not a maintainer here. If you could merge that'd be great!

dnovillo:
Done. Thanks for the fix!

PR #514, opened 2025-05-07
Register Goopax SPIR-V generator

josopait:
I would like to register our Goopax Vulkan backend SPIR-V generator, which we will introduce with our upcoming release.


bashbaug:
Merging as discussed in the May 7th teleconference.

PR #524, opened 2025-06-17
Use original LICENSES/CC-BY-4.0.txt

dneto0:
This is needed by REUSE

### DIFF
--- a/LICENSES/CC-BY-4.0.txt
+++ b/LICENSES/CC-BY-4.0.txt
@@ -1,396 +1,156 @@
-Attribution 4.0 International
+Creative Commons Attribution 4.0 International
 
-=======================================================================
-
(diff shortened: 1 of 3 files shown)


KhronosWebservices:
I should add, if there are any security concerns about using reuse-compliance-check, we do have this GitHub Org where we can add verified actions: https://github.com/KhronosGroupActions

Naghasan:
> Could/should we add some sort of REUSE checks to our CI?

Added a job using the same recipe as VK https://github.com/KhronosGroup/SPIRV-Headers/pull/525

outofcontrol:
@dneto0 Correct, adding the SPDX header to the file for license and copyright is fine too. And technically we don't need to have a REUSE.toml file if all the files have correct headers.

bashbaug:
Merging (offline) as discussed in the June 25th teleconference.

PR #537, opened 2025-07-28
Fixed typo in operand name of OpSubgroupAvcImeSetDualReferenceINTEL

boredhuman:
Was making some tooling and noticed OpSubgroupAvcImeSetDualReferenceINTEL seemed to accidently include id> in one of the operand names. Search Window Config is also the operand name for two other instructions both of which don't seem to include the id> prefix.


bashbaug (approved):
Thank you for fixing this!

PR #542, opened 2025-08-20
Update SPV_INTEL_variable_length_array to rev 3

MrSidims:
Spec: https://github.com/KhronosGroup/SPIRV-Registry/pull/352

### DIFF
--- a/include/spirv/unified1/spirv.core.grammar.json
+++ b/include/spirv/unified1/spirv.core.grammar.json
@@ -9159,7 +9159,7 @@
       "operands" : [
         { "kind" : "IdResultType" },
         { "kind" : "IdResult" },
-        { "kind" : "IdRef", "name" : "Lenght" }
+        { "kind" : "IdRef", "name" : "Length" }
       ],
       "capabilities" : [ "VariableLengthArrayINTEL" ],
       "version" : "None"
@@ -10846,7 +10846,7 @@
         { "kind" : "IdResultType" },
         { "kind" : "IdResult" },
         { "kind" : "IdRef", "name" : "Element Type" },
-        { "kind" : "IdRef", "name" : "Lenght" }
+        { "kind" : "IdRef", "name" : "Length" }
       ],
       "capabilities" : [ "UntypedVariableLengthArrayINTEL" ],
       "version" : "None"


include/spirv/unified1/spirv.core.grammar.json
```diff
@@ -17735,6 +17747,13 @@
           "extensions" : [ "SPV_INTEL_ternary_bitwise_function"],
           "version" : "None"
         },
+        {
+          "enumerant" : "UntypedVariableLengthArrayINTEL",
+          "value" : 6244,
```
bashbaug, line 17752: In the spec PR https://github.com/KhronosGroup/SPIRV-Registry/pull/352 this capability has value 6243, not 6244.

```suggestion
          "value" : 6243,
```

include/spirv/unified1/spirv.core.grammar.json
```diff
@@ -10838,6 +10838,18 @@
       "capabilities" : [ "TernaryBitwiseFunctionINTEL" ],
       "version" : "None"
     },
+    {
+      "opname" : "OpUntypedVariableLengthArrayINTEL",
+      "class"  : "@exclude",
```
bashbaug, line 10843: This is consistent with the rest of the variable length array instructions, but I think we should remove `@exclude` now and add all of the variable length array instructions to some other class instead (maybe Memory?).  Rationale: There aren't very many variable length array instructions, and the extension will be published on the registry shortly.

Naghasan: +1, I don't think exclude is right (unless you really want to hide this in the core spec). Memory is more appropriate (and that's what the extension mentions anyway).

include/spirv/unified1/spirv.core.grammar.json
```diff
@@ -10838,6 +10838,18 @@
...
+      "opcode" : 6244,
+      "operands" : [
+        { "kind" : "IdResultType" },
+        { "kind" : "IdResult" },
+        { "kind" : "IdRef", "name" : "Lenght" }
```
Naghasan, line 10848:
```suggestion
        { "kind" : "IdRef", "name" : "Element Type" },
        { "kind" : "IdRef", "name" : "Length" }
```

Naghasan: `OpVariableLengthArrayINTEL` also has the typo

MrSidims: Thanks! Though I'm not sure, if I see a typo in OpVariableLengthArrayINTEL.

include/spirv/unified1/spirv.core.grammar.json
MrSidims: Applied, thanks!

include/spirv/unified1/spirv.core.grammar.json
MrSidims: ah, thanks! my classical mistake, hope I'll learn how to spell it correctly one day

bashbaug (approved):

PR #574, opened 2026-03-23
Add SPV_KHR_constant_data

r-potter:


include/spirv/unified1/spirv.core.grammar.json
```diff
@@ -15131,6 +15155,12 @@
           ],
           "version" : "None"
         },
+        {
+          "enumerant" : "UTFCodePointsKHR",
```
bashbaug: @Tobski in the spec this is named "UTFEncodedKHR".

https://github.khronos.org/SPIRV-Registry/extensions/KHR/SPV_KHR_constant_data.html#_modifications_to_the_spir_v_specification_version_1_6_4

Which is correct?

Tobski: It got renamed in the spec during development and I imagine we just didn't update it here, so it should be "UTFEncodedKHR".

r-potter: @Tobski @bashbaug Updated

bashbaug (approved):
LGTM

PR #584, opened 2026-04-23
Update headers for spirv issue 373

alelenv:
Issue : https://github.com/KhronosGroup/SPIRV-Registry/issues/373


bashbaug (approved):
Merging as discussed in the April 29th teleconference.

PR #586, opened 2026-04-29
Reserve some tokens for Samsung

mmoult:


mmoult:
@r-potter As discussed

bashbaug:
Merging as discussed in the April 29th teleconference.

As a reminder, the reserved block is only for non-instruction enums, so it's good for things like new capabilities or decorations, but if new instructions are needed then they will need another dedicated enum reservation.

PR #587, opened 2026-04-30
Fix OpCompositeConstructContinuedINTEL operands

aobolensk:
Remove IdResultType and IdResult from OpCompositeConstructContinuedINTEL

This instruction is a continuation that supplies additional constituents to a preceding OpCompositeConstruct and does not produce its own result

### DIFF
--- a/include/spirv/unified1/spirv.core.grammar.json
+++ b/include/spirv/unified1/spirv.core.grammar.json
@@ -11223,8 +11223,6 @@
       "class"  : "Composite",
       "opcode" : 6096,
       "operands" : [
-        { "kind" : "IdResultType" },
-        { "kind" : "IdResult" },
         { "kind" : "IdRef",        "quantifier" : "*", "name" : "Constituents" }
       ],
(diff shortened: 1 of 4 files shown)


aobolensk:
@MrSidims please take a look

MrSidims (approved):
Good catch!

FYI @bashbaug I believe unified spec should be corrected, see
https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#OpCompositeConstructContinuedINTEL
vs definition in
https://github.khronos.org/SPIRV-Registry/extensions/INTEL/SPV_INTEL_long_composites.html

bashbaug (approved):
Thank you for fixing this.

bashbaug:
> I believe unified spec should be corrected [...]

Thanks, yeah, the unified spec is generated from the SPIR-V grammar file, so this fix should propagate automatically into the next version of the spec.

bashbaug:
Vendor extension fix, no reason to wait to merge this one.

PR #591, opened 2026-05-01
Add NonSemantic.Graph.DebugInfo instruction set

kpet:


kpet:
Specification: https://github.com/KhronosGroup/SPIRV-Registry/pull/406

bashbaug (approved):
Merging as discussed in the May 13th teleconference.

PR #596, opened 2026-05-06
Spv qcom image processing3

wooyoungqcom:


alan-baker (approved):

alan-baker:
Something is wrong with your latest rebuild. It is no longer including the new enum when it tests the build.

wooyoungqcom:
I am having a CLA issue. Maybe that is the reason the server refuses to pull in the patch?

include/spirv/unified1/spirv.core.grammar.json
```diff
@@ -5408,6 +5408,22 @@
...
+      "opname" : "OpImageGatherQCOM",
+      "class"  : "Image",
+      "opcode" : 4545,
+      "operands" : [
+        { "kind" : "IdResultType" },
+        { "kind" : "IdResult" },
+        { "kind" : "IdRef", "name" : "Sampled Image" },
+        { "kind" : "IdRef", "name" : "Coordinate" },
+        { "kind" : "IdRef", "name" : "Component" },
+        { "kind" : "LiteralInteger", "name" : "Mode" },
```
alan-baker: I think this should be:

```suggestion
        { "kind" : "GatherModes", "name" : "Mode" },
```

alan-baker: Separately, looking at the spec again. The extension claims this should be an id.

alan-baker (changes requested):

include/spirv/unified1/spirv.core.grammar.json
wooyoungqcom: It should have been GatherModes. Also, while I was fixing it, I realized I missed adding a couple of files to the change set. That mistake also has been fixed. Thank you so much !!!

alan-baker (approved):

include/spirv/unified1/spirv.core.grammar.json
wooyoungqcom: We made the changes quite a while ago, and I was confused with the back-up changes I made.
The intention was we use <id> for "Mode". 
I made the fix and I also updated the spec accordingly.

wooyoungqcom:
@alan-baker Could you kindly merge the pull request? thanks much !!!

PR #472, opened 2025-01-13
Add a source language for Rust

wingertge:
Adds source language identifiers for [Rust](https://www.rust-lang.org/). There's an experimental compiler backend for SPIR-V ([rust-gpu](https://github.com/Rust-GPU/rust-gpu)), as well as a mostly stable SPIR-V backend as part of [cubecl](https://github.com/tracel-ai/cubecl).

PR #606, opened 2026-05-25
Add Pred as a new SPIR-V source language

JebScape:


kpet:
These headers are generated. You need to modify https://github.com/KhronosGroup/SPIRV-Headers/blob/main/include/spirv/unified1/spirv.core.grammar.json. See https://github.com/KhronosGroup/SPIRV-Headers/pull/472 for an example.

kpet:
Also, please include a link to a page describing the Pred language if one exists. A quick search does not seem to return relevant results.

JebScape:
> These headers are generated. You need to modify https://github.com/KhronosGroup/SPIRV-Headers/blob/main/include/spirv/unified1/spirv.core.grammar.json. See #472 for an example.

Okay thanks!



> Also, please include a link to a page describing the Pred language if one exists. A quick search does not seem to return relevant results.

No public information exists aside from some comments I have made on X. It is a safety-critical shader language that we have been developing at Isochron Technologies, LLC. It is designed to guarantee fixed execution times, thereby making it trivial to verify that worst-case execution times will be met independent of data inputs (by always running in the worst case at all times).

We have a team of 3 people working on it at this time.

JebScape:
I believe that's everything now.

alan-baker:
You may want to also register a generator in the spir-v.xml.

JebScape:
> You may want to also register a generator in the spir-v.xml.

Done.

alan-baker (approved):

bashbaug:
Merging as discussed in the May 27th teleconference.

PR #611, opened 2026-06-30
Fix Int64ImageEXT capability prerequisite to Int64

aobolensk:
The Int64ImageEXT capability listed Shader as its prerequisite, but per SPV_EXT_shader_image_int64 spec (https://github.khronos.org/SPIRV-Registry/extensions/EXT/SPV_EXT_shader_image_int64.html) it implicitly declares Int64

### DIFF
--- a/include/spirv/unified1/spirv.core.grammar.json
+++ b/include/spirv/unified1/spirv.core.grammar.json
@@ -17676,7 +17676,7 @@
         {
           "enumerant" : "Int64ImageEXT",
           "value" : 5016,
-          "capabilities" : [ "Shader" ],
+          "capabilities" : [ "Int64" ],
           "extensions" : [ "SPV_EXT_shader_image_int64" ],
           "version" : "None"
         },


aobolensk:
@MrSidims please take a look

include/spirv/unified1/spirv.core.grammar.json
```diff
@@ -17676,7 +17676,7 @@
         {
           "enumerant" : "Int64ImageEXT",
           "value" : 5016,
-          "capabilities" : [ "Shader" ],
+          "capabilities" : [ "Int64" ],
```
MrSidims, line 17679: Probably should be rather 
```suggestion
          "capabilities" : [ "Shader, Int64" ],
```

MrSidims: Nvm, should be just `Int64`, but would be nice to hear from @Tobski

MrSidims (approved):

include/spirv/unified1/spirv.core.grammar.json
Tobski: Looks like this was just a mistake in the transcribe to SPIR-V headers, so this is correct.

Tobski (approved):

bashbaug:
Merging as discussed in the July 8th teleconference.

PR #617, opened 2026-07-27
switch to the canonical CC-BY-4.0 license text

### LINKED ISSUE #616: CC-BY-4.0.txt does not use the canonical license text
PR #524 changed the format of `CC-BY-4.0.txt` so that it does not match the [canonical text](https://creativecommons.org/licenses/by/4.0/legalcode.txt). The PR mentions this is needed by REUSE, but the formatting breaks other license checkers that expect the licenses to be byte-identical. I'm not familiar with REUSE, but I imagine there is a way to make it happy with the unmodified text.

bashbaug:
Fixes #616 

Specifically, this uses the exact license text from: https://creativecommons.org/licenses/by/4.0/legalcode.txt

The only changes I can see are:

1. Whitespace changes (expected, although this does make a diff challenging).
2. URLs in a few places, e.g. "More considerations for licensors: wiki.creativecommons.org/Considerations_for_licensors"
3. Small editorial changes, e.g. dash style, list capitalization, etc.

### DIFF
--- a/LICENSES/CC-BY-4.0.txt
+++ b/LICENSES/CC-BY-4.0.txt
@@ -1,156 +1,395 @@
-Creative Commons Attribution 4.0 International
+Attribution 4.0 International
 
- Creative Commons Corporation (“Creative Commons”) is not a law firm and does not provide legal services or legal advice. Distribution of Creative Commons public licenses does not create a lawyer-client or other relationship. Creative Commons makes its licenses and related information available on an “as-is” basis. Creative Commons gives no warranties regarding its licenses, any material licensed under their terms and conditions, or any related information. Creative Commons disclaims all liability for damages resulting from their use to the fullest extent possible.
+=======================================================================
+
+Creative Commons Corporation ("Creative Commons") is not a law firm and
+does not provide legal services or legal advice. Distribution of
+Creative Commons public licenses does not create a lawyer-client or
+other relationship. Creative Commons makes its licenses and related
+information available on an "as-is" basis. Creative Commons gives no
+warranties regarding its licenses, any material licensed under their
+terms and conditions, or any related information. Creative Commons
+disclaims all liability for damages resulting from their use to the
+fullest extent possible.
 
 Using Creative Commons Public Licenses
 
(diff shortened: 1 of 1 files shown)


bashbaug:
Note, the CI job to "Verify repository license compliance" passed, so it seems that REUSE is happy with this license text.

LICENSES/CC-BY-4.0.txt
```diff
@@ -1,156 +1,395 @@
...
+=======================================================================
+
+Creative Commons is not a party to its public
+licenses. Notwithstanding, Creative Commons may elect to apply one of
+its public licenses to material it publishes and in those instances
+will be considered the “Licensor.” The text of the Creative Commons
```
dneto0, line 381: This sentence has been added:

> The text of the Creative Commons
public licenses is dedicated to the public domain under the CC0 Public
Domain Dedication.

That reads as only affecting the text of the license itself. That seems fine.

dneto0 (approved):
I've confirmed that almost all the changes are as you noted:

- punctuation
- different codepoints for hypens
- sub-bullet numbering
- line wrap
- addition or subtraction of empty lines and only-equals lines

There is one textual addition noted below.  It seems fine to me as it only applies to the license text itself.

bashbaug:
Merging as discussed in the July 29th teleconference.

LICENSES/CC-BY-4.0.txt
oddhack: N.b. 'wdiff -3' is helpful for this sort of thing.

REUSE does not AFAIK care what the contents of the files in LICENSES are, only that they exist. It's on the repo owner to ensure the licenses actually contain the right words. There is utility in keeping up to date but I don't think we are under any obligation to be "byte for byte identical" to a specific version of a license at a specific point in time for the hypothetical benefit of other compliance tools that we don't use ourselves. Licenses as published by the originators can and do get tweaked in minor ways over time and downstream users aren't obligated to track every such tweak. Khronos has done this with our own licenses, too.

PR #620, opened 2026-08-01
Correct name and comment

metora-xl:
Fix the name that I misunderstood the Mesa3D GLSL-IR, and change the comment to correct GitHub URL.

The Mesa-IR is converted from GLSL-IR.
The Mesa-IR is old and deprecated and removed since Mesa3D v22.0.0.


bashbaug (approved):
Merging as discussed in the August 5th teleconference.

PR #621, opened 2026-08-04
Add SPV_EXT_cooperative_matrix_maintenance1

jeffbolznv:
Spec is at https://github.com/KhronosGroup/SPIRV-Registry/pull/443.


include/spirv/unified1/spirv.core.grammar.json
```diff
@@ -6407,15 +6407,16 @@
...
+      "aliases" : ["OpCooperativeMatrixConvertNV"],
       "class"  : "Conversion",
       "opcode" : 5293,
       "operands" : [
         { "kind" : "IdResultType" },
         { "kind" : "IdResult" },
         { "kind" : "IdRef",             "name" : "Matrix" }
       ],
-      "capabilities" : [ "CooperativeMatrixConversionsNV" ],
+      "capabilities" : [ "CooperativeMatrixConversionsEXT", "CooperativeMatrixConversionsNV" ],
```
alan-baker, line 6419: Nit: I'd just drop the NV version as you did below for per-element ops.

jeffbolznv: Most of the NV/EXT capabilities alias, but this one does not (there were some changes to transpose behavior and supported uses). So this should stay as-is.

I don't remember why we didn't change the opcode enum at the same time. I imagine we discussed it, but I don't remember any details.

alan-baker (approved):

PR #629, opened 2026-09-02
Tags use a combined SemVer CalVer scheme

### LINKED ISSUE #624: versioning scheme that is friendly to SemVer-based systems?
As the README describes, this repo has two tagging schemes:
- tags based on the Vulkan SDK releases.  (scheme `vulkan-sdk-1.X.YYY.Z`)
-  a plan for CalVer-based numbers: `vYYYY.0M.R`

**What is the recommended way to incorporate SPIRV-Headers in a SemVer -based system?**

This came up in real life.  The Bazel build ecosystem now has a Bazel Central Registry (https://registry.bazel.build/ ) where people publish module sources using semver-based numbering.   Bazel modules declare their dependencies with version numbers, and the build system resolves conflicts using SemVer assumptions, e.g. that it's harmless to upgrade from 1.4 to 1.5.

Someone published a [`spirv_headers` module](https://registry.bazel.build/modules/spirv_headers) numbered in the vulkan-sdk way.  i..e. `1.4.341.0`.    That probably works great for the Vulkan ecosytem, but it doesn't seem quite canonical.

Now that CalVer tagging has arrived, we could use that.  But it doesn't play too nicely with SemVer because of the artificial jump from one year to the next.  

The options include:
1. rely on the vulkan numbering only.   Seems non-canonical
2.  don't solve this problem, use CalVer only
3. prefix CalVer with a `1.` forever since we promise backward compatibility for the foreseeable future.
     - For example, we get `1.2026.08.0`
     - And if we actually break backward compatibility, then we can change the leading 1 to 2, etc.

(I assume the working group has already decided against using SPIR-V version numbers, e.g. 1.6.7.x)

**I lean toward option 3.** 

That does not mean we have to tag the repo any differently.   The semver numbering would be a *rule* that is applied when making semver-numbered artifacts.  E.g. a Bazel Central Registry for `v2026.08` would use "1.2026.08" in the module definition ([here](https://github.com/KhronosGroup/SPIRV-Headers/blob/main/MODULE.bazel#L7)).

dneto0:
Tag format

   v1.YYYY0M.R

represents semantic version:

   1.YYYY0M.R

where YYYY is a four-digit year, 0M is a zero-padded two digit month, and R is a non-negative integer for distinguishing releases within a month.

Note: Strict SemVer only allows three numeric components (MAJOR.MINOR.PATCH), and disallows zero-padded numeric compoentns (which makes a zero-padded month field problematic when on its own).

Fixed: #624

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -4,7 +4,7 @@
 
 module(
     name = "spirv_headers",
-    version = "0.0.0",
+    version = "1.202609.0-pre.0",
 )
 
 bazel_dep(name = "rules_cc", version = "0.1.1")
--- a/README.md
+++ b/README.md
@@ -209,10 +209,29 @@ python3 bin/makeExtinstHeaders.py
      to record the SPIR-V headers used by that version of the Vulkan SDK.  For
      example, the tag `vulkan-sdk-1.X.YYY.Z` indicates the SPIR-V headers that
      were included in the Vulkan SDK version `1.X.YYY.Z`.
-  2. For other uses, the repo is tagged using a [CalVer](https://calver.org/)
-     date-based scheme on an as-needed basis.  For example, the tag `vYYYY.0M.R`
-     indicates release `R` of the SPIR-V headers in the zero-padded month `0M`
-     of the full year `YYYY`.
+  2. For other uses, the repo is tagged on an as-needed basis using a scheme
+     that uses both [Semantic Versioning](https://semver.org)
+     and [Calendar Versioning](https://calver.org) conventions.
+
+     The tag is of the form:
+
+         v1.YYYY0M.R
+
+     representing version number `1.YYYY0M.R`:
+
+     * The first numbered component is `1`:
+         * This fits with Semantic Versioning because updates should always be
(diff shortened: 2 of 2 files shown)


dneto0:
This is not exactly as discussed in the working group, so it needs a review again.

MODULE.bazel
```diff
@@ -4,7 +4,7 @@
 
 module(
     name = "spirv_headers",
-    version = "0.0.0",
+    version = "1.202609.0-pre.0",
```
dneto0, line 7: And, for example, if a `1.202609.0` is finalized, then the `-pre.0` suffix can be stripped with a patch when registering it with the BCR.

dneto0 (commented):
This patch also includes an update to the bazel module version, with a prerelease suffix.

bashbaug (approved):

bashbaug:
Merging as discussed in the September 9th teleconference.

We still need to push a tag following this convention.

bashbaug:
I just pushed a tag following this convention.  Please check that this looks correct - thanks!

https://github.com/KhronosGroup/SPIRV-Headers/releases/tag/v1.202609.0

PR #630, opened 2026-09-09
Start 1.202609.1 or later

dneto0:
v1.202609.0 has been tagged

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,7 +7,9 @@
 #    https://www.khronos.org/registry/
 
 cmake_minimum_required(VERSION 3.14)
-project(SPIRV-Headers LANGUAGES C CXX VERSION 1.5.5)
+project(SPIRV-Headers
+	LANGUAGES C CXX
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -7,7 +7,9 @@
 #    https://www.khronos.org/registry/
 
 cmake_minimum_required(VERSION 3.14)
-project(SPIRV-Headers LANGUAGES C CXX VERSION 1.5.5)
+project(SPIRV-Headers
+	LANGUAGES C CXX
+	VERSION 1.202609.1)
```
dneto0, line 12: The CMake docs say this only accepts non-negative number components.  No `pre`-like suffixes.

dnovillo (approved):
LGTM

bashbaug (approved):
LGTM, but for my own reference if nothing else: When should we bump these values in these files? Before tagging? After tagging? Both?

It'd be nice if this were written down somewhere since I know I'll forget otherwise.  Thanks!

PR #631, opened 2026-09-09
Add minimal tests

dneto0:
Enable testing if this is the top-level project. Then "ctest" will run the tests.

Builds on #630 but is not semantically tied to it.


bashbaug (commented):
Should we update CI to run these tests also?  Our "presubmit" step builds the tests, but it doesn't run them.

PR #633, opened 2026-09-10
fix enabling capability for generic pointer instructions

bashbaug:
Specifically, this PR fixes the enabling capability for the instructions:

* OpGenericPtrMemSemantics, OpPtrCastToGeneric, OpGenericCastToPtr, and OpGenericCastToPtrExplicit.

It is more accurate to gate these instructions by GenericPointer rather than Kernel, since some Kernel SPIR-V consumers may not support generic pointers, and Shader SPIR-V consumers could conceivably support generic pointers in the future.

Note that these instructions all operate on pointers that point to the generic storage class, so the GenericPointer capability was already required implicitly.

fixes internal issue 784


alan-baker (approved):

bashbaug:
Discussed in the September 16th teleconference. I'll check SPIRV-Tools before merging.

bashbaug:
> I'll check SPIRV-Tools before merging.

I've verified that SPIRV-Tools builds and all tests pass with these changes.  Notably, all of the tests I found that test the generic address space instructions listed in the PR description already declare **GenericPointer** as part of `GenerateKernelCode()`, as expected, since they also declare pointers to the **Generic** storage class.

bashbaug:
Merging (offline) as discussed in the September 16th teleconference.

PR #636, opened 2026-09-19
fix TaskPayloadWorkgroupEXT storage class version

### LINKED ISSUE #632: SPV_EXT_mesh_shader: TaskPayloadWorkgroupEXT storage class and SPIR-V 1.4
The SPIR-V grammar file currently lists SPIR-V 1.4 as a "version" for the TaskPayloadWorkgroupEXT storage class:

https://github.com/KhronosGroup/SPIRV-Headers/blob/04fd3caa1e8267e4d95c806cad901181728e1006/include/spirv/unified1/spirv.core.grammar.json#L13897-L13903

In this context, I believe the "version" is used to describe when the enumerant was added to a core SPIR-V version.  Since SPV_EXT_mesh_shader and this storage class is still exclusive to an extension, though, I think that the "version" is incorrect, and this should be changed to "None" instead.

Note that the other new enumerants added by [SPV_EXT_mesh_shader](https://github.khronos.org/SPIRV-Registry/extensions/EXT/SPV_EXT_mesh_shader.html), such as the TaskEXT and MeshEXT execution models, list "None" as their version in the grammar.

Is there a reason to list SPIR-V 1.4 for the TaskPayloadWorkgroupEXT storage class, or should this be changed to "None" instead?

Aside: The [Machine Readable Grammar](https://github.com/KhronosGroup/SPIRV-Docs/blob/main/specs/MachineReadableGrammar.adoc) document describing the JSON schema is now available in the SPIR-V docs repo, but it doesn't currently describe the "version".  We should fix this, too.

bashbaug:
The TaskPayloadWorkgroupEXT storage class is only part of the SPV_EXT_mesh_shader extension and has not been promoted to a core SPIR-V version, so it should have version set to None, not 1.4.

fixes #632


bashbaug:
I did a quick check and did not observe any SPIRV-Tools issues with this change.

alan-baker (approved):

bashbaug:
Merging as discussed in the September 23rd teleconference.

